PR #8, opened 2017-06-16
SonarQube Endpoint

miiitch:
The Sonarqube Endpoint is a like a generic endpoint with the type "sonarqube" We can easily add a generic endpoint from this base but I think a refactoring should be welcomed to avoid code duplication.


DarqueWarrior:
This cmdlet requires that you install an extension. It just so happens on my test account I don't have it installed and you get a less that useful error message.  Can you uninstall the extension from an account and test?  That way you can add better error handling to tell the user which extension they need to install.  At a minimum that needs to be in the help.

Second suggestion is make the token secured if entered from the prompts.  Look at the Add-TeamAccount as an example. There you can provide the PAT two ways. One way as plain text good for use in scripts and one way as a Secure String good for use of interactive sessions.

miiitch:
So:
- I need to test the code without the extension to provide a meaningful message if the extension is not installed
- I need to use a SecureString for the token inside a parameter set

DarqueWarrior:
Correct. With the token you should be able to pass it in either as a plain text as a command line arg or as a secure string if PowerShell prompts for it. Add-TeamAccount does this with the PAT.

miiitch:
Hi @DarqueWarrior, better?

PR #10, opened 2017-07-19
Adding a regular expression to validate VSTS account

eulesv:
A simple fix in case someone provides an account without protocol (account.visualstudio.com) or if the name includes http (for example myhttpaccount as account name)


DarqueWarrior:
Sorry for delay. I will review soon.

PR #15, opened 2017-09-05
Add support for teams

KeesV:
- Added support for creating / deleting / updating teams
- Added support for getting team members

### DIFF
--- a/Team.psd1
+++ b/Team.psd1
@@ -106,7 +106,7 @@
       'Get-ServiceEndpoint',
       'Get-TeamInfo',
       'Get-Team',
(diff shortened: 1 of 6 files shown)


DarqueWarrior:
When I started testing I called Get-Team. After that worked. I ran Get-Team | Get-TeamMembers and I got a lot of errors.  I would expect this to work.

DarqueWarrior:
Oh by the way. I like your update! :)

KeesV:
I kind of forgot to properly implement pipeline support... Should work now, so you can do things like
`Get-Project | Get-Team | Get-TeamMember`

PR #17, opened 2017-09-14
Add ProjectName as a property on team member so it can be used further down the pipeline

KeesV:
Add ProjectName as a property on team member so it can be used further down the pipeline.

For example:
`Get-Project | Get-Team | Get-TeamMember | select projectName,team,displayName | Sort-Object -Property ProjectName,Team,displayName`
This now gives a useful overview of team members in each team project / team in an account.


DarqueWarrior:
I plan to publish this weekend. Thanks for the continued support.

PR #25, opened 2017-10-10
Add function to get user account details

KeesV:
Adds a function to get user account details, which are found under "Users" at the account level in the GUI.
This makes it possible to create automated usage / licensing reports. For example:
`Get-Project | Get-Team | Get-TeamMember | % { $teammember = $_; $account = Get-VSAccountUser $_.uniqueName; New-Object -TypeName PSObject -Property @{ Project = $teammember.ProjectName; LastAccessed = $account.LastAccessed; License = $account.LicenseType; Team = $teammember.Team; DisplayName = $teammember.DisplayName } } | Format-Table`


DarqueWarrior:
Name the function Get-VSTeamAccountUser you can add an alias for Get-VSAccountUser if you like. Also please add unit tests. Otherwise, looks good.

KeesV:
Thanks for the comments! Should be good now

DarqueWarrior:
I think this only works in VSTS. How are we going to handle the fact this module supports TFS as well?

KeesV:
Good point. I think there's no sensible equivalent in TFS. We could extract
some data from Access Levels, but I think that should be in a different
function.
Maybe just throw an error when we're on TFS?

DarqueWarrior:
You can use the Get-VSTeamOption function to see if the account we are connected to supports the API. If not I don't think I want to throw but it should display a message stating this call is not supported on your account.

KeesV:
Since this is a call to <account>.vssps.visualstudio.com, I don't think we can use Get-VSTeamOption here, since that calls <account>.visualstudio.com to get the available API's. There doesn't seem to be an equivalent call to <account>.vssps.visualstudio.com (at least I couldn't find it). 
For now I solved it by writing a very simple _isVSTS function that checks if the account URL ends in ".visualstudio.com". This is now in the usermanagement.psm1 file, but we might want to move this to somewhere more generic. What do you think?

DarqueWarrior:
I reviewed the code and made some changes. I also merged with latest version from master.

I have one final ask. Please add help in the .docs folder. It is simply markdown now. Just copy and paste similar topic and update. You will need file in .docs and .docs/synopsis.  I will merge once help is written. 

Nice work and thanks for the continued support.

DarqueWarrior:
I am working on adding users management using the newly released Entitlements API.

PR #26, opened 2017-10-17
Remove deadlock when endpoint creation failed

miiitch:
If and endpoint creation failed, the status of the endpoint is kept at $false and the _trackProgress method never ends.

If the status is $false, I add a check on the operational status to verify if the creation is still in progress. If there is a failure I write the failure message on the error output


DarqueWarrior:
Sorry for the delay I will try and review today.

DarqueWarrior:
Please add unit tests to cover your new code.

miiitch:
You're right, it's better. Is it consistant with the other parts of the module?

DarqueWarrior:
The only other place where I track a process is creating a project. I want to complete the PR then we can see if we want to add that there.

PR #39, opened 2018-02-20
Fix GET Approval filter by release ids

franciscocabral:
API [documentation](https://www.visualstudio.com/en-us/docs/integrate/api/rm/approvals) says that to filter by release id is needed the query param `releaseIdsFilter` not `releaseIdFilter`


DarqueWarrior:
Thanks for spotting this. I will update the help to match. I will also add an alias for the parameter so we don't break any scripts that might already be setting that value. Finally I will bump the version.

PR #42, opened 2018-03-14
Update-VSTeamBuildDefinition

karpis:
Added a function for updating an existing build definition from an input json file.


DarqueWarrior:
Reviewing today.

DarqueWarrior:
That was one of the best PRs I have ever received. Great job and thanks.

PR #46, opened 2018-04-04
Build sourcebranchname

karpis:
Added optional -SourceBranch parameter to Add-VSTeamBuild. Setting this overrides the default source branch set in the build definition for the build.


karpis:
@DarqueWarrior There was no hurry. By the way, when will 2.1.10 be published to the PowerShell Gallery?

PR #66, opened 2018-07-26
adapt integration test to additional VSTS pool 'Hosted Ubuntu 1604'

KaiWalter:
# PR Summary

VSTS now has 6 agent pools / queues which breaks the integration test


DarqueWarrior:
I will be working on vsteam tomorrow.

PR #68, opened 2018-07-29
added drilldown on BuildDefinition and added integration test

KaiWalter:
# PR Summary

- added more type mapping on Get-VSTeamBuildDefinition so that Phases and Target can be explored more easily

- added an integration test for BuildDefinition

Right know I put type info related to Process / Phase entities into 1 `process.ps1xml`. Let me know whether I can carry on with that approach or if you want to have one `.ps1xml` for each entity.


KaiWalter:
@DarqueWarrior can you please also stage and commit types\pools.ps1xml
then I would merge upstream and resubmit the PR

DarqueWarrior:
I moved pools.ps1xml when I added the classes to the provider. If you review Get-VSTeamPool and Get-VSTeamAgent you will see I no longer apply types and I instead return instances of the classes.  As I move entities into the provider I add the classes and remove the type files. With the classes the type files are redundant. 

Until there are classes I would prefer there be one ps1xml for each type. I believe that is how I have done them all up to this point. 

I will however, be reducing the number of format files.

KaiWalter:
Good thinking. OK, but then let me also move my extensions into classes. This will help later when VSTeam needs to support Update on Process / Phase / Step.

PR #71, opened 2018-08-01
Integration tests for Build Definitions

KaiWalter:
# PR Summary

Added integration tests for Build Definitions:
- adding one build definition with 1 and one with 2 phases


DarqueWarrior:
I have to update the test for TFS2017 and TFS2018. It is failing at the moment. It might not be supported there. I need to do some research

PR #82, opened 2018-08-22
added GitRepository object to BuildDefinition

KaiWalter:
# PR Summary

- removed output of objects returned from API to avoid misleading outputs for consumers
- added GitRepository object to BuildDefinition
- had to add a ```[VSTeamProjectCache]::timestamp = -1``` to ```VSTeamAccount``` constructor after experiencing blocking with ```New-PSDrive -Name V -PSProvider SHiPS -Root 'VSTeam#VSTeamAccount' -Verbose; Get-VSTeamBuildDefinition -ProjectName someProject```

## PR Checklist


KaiWalter:
aw, sorry - now I see it and will fix it

KaiWalter:
> again a learning - no matter how good your intentions are and how insignificant a change seems: never skip the unit tests

Now unit and integration tests for build definitions work again.

The problem I was trying to solve is that with the API call for ```Get-VSTeamBuildDefinition``` a repository object is returned which is missing ```remoteURL``` - hence I thought getting my own object. I need to into this again.

KaiWalter:
@DarqueWarrior as a next step I'd like to convert Queue also to class VSTeamQueue in teamspsdrive. My end game is that I can conveniently replace/exchange/update or even add elements in a build definitions tree and then update the whole definition. Hence I think each elements needs to have a proper class so it can be instantiated by the consumer.

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ Build Definition / Process / Phases / Steps are broken down into separate object
 (Get-VSTeamBuildDefinition -ProjectName MyProject -Id 42).Process.Phases[0].Steps[0]
 ```
 
+<<<<<<< HEAD
```
nschonni: Looks like a merge issue snuck in

PR #93, opened 2018-09-13
replaced accountname.visualstudio.com with dev.azure.com/accountname

KaiWalter:
# PR Summary

- replaced accountname.visualstudio.com with dev.azure.com/accountname
- aligned subdomain build url
- aligned tests
- executed all units and integration tests successfully (for VSTS)

## PR Checklist


DarqueWarrior:
You have to make sure this code still works with old profiles.  For example, if I already have a profile called 'test' and the URL is 'test.vsiualstudio.com'.  If I were to try and run *Add-Profile* for 'test' again the original entry should be updated to 'dev.azure.com/test' but instead, I get two entries both named 'test' with two different URLs.

Instead of changing the 'test.visualstudio.com' entry to 'dev.azure.com/test' in the $contents var in profiles.Tests.ps1 you needed to add a new entry leaving the old one as well. Then test updating them both to make sure old profiles can still be used.

KaiWalter:
OK, I see. I look into this.

KaiWalter:
@DarqueWarrior handled old URLs in profile and added unit tests to verify the conversion from old to new in profile

PR #128, opened 2019-01-17
Add Process Template Support

fifthstreetpartners:
# PR Summary

## PR Checklist

- [x] [Update VSTeam.psd1](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#add-a-format-file)
- [x] [Update CHANGELOG.md](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#add-a-format-file)

### DIFF
--- a/.docs/Add-VSTeamProject.md
+++ b/.docs/Add-VSTeamProject.md
@@ -47,15 +47,15 @@ Position: 0
 
 ### -ProcessTemplate
 
-The name of the process template to use for the project. The acceptable values for this parameter are:
+The name of the process template to use for the project.
 
-- Agile
-- Scrum
-- CMMI
+You can tab complete from a list of available projects.
+
+Will use the default process template if one is not provided
 
 ```yaml
 Type: String
-Default value: Scrum
+Default value: [Determined by Default Process Template definition]
 ```
 
 ### -Description
@@ -85,3 +85,5 @@ Type: SwitchParameter
 [Add-VSTeamAccount](Add-VSTeamAccount.md)
 
 [Remove-VSTeamProject](Remove-VSTeamProject.md)
+
+[Get-VSTeamProcess](Get-VSTeamProcess.md)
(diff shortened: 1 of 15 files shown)


DarqueWarrior:
The alias Get-Process is going to cause problems.  I would suggest you just don't add it. Get-Process is an out of the box PowerShell cmdlet. It is very popular and we don't want to take that one over.

DarqueWarrior:
My only other feedback is what should be the default format of the results to Get-VSTeamProcess? Currently it is a list and it shows the ID and URL. I am not sure why anyone would need to see that by default. I might suggest adding a table format that only shows the name and description. Users could always run Get-VSTeamProcess | FL * to get the current list view.

This is really nice work.

fifthstreetpartners:
> 
> 
> The alias Get-Process is going to cause problems. I would suggest you just don't add it. Get-Process is an out of the box PowerShell cmdlet. It is very popular and we don't want to take that one over.

I thought about that actually but figured I'd include it just in case.  No problem taking it out.

fifthstreetpartners:
> 
> 
> My only other feedback is what should be the default format of the results to Get-VSTeamProcess? Currently it is a list and it shows the ID and URL. I am not sure why anyone would need to see that by default. I might suggest adding a table format that only shows the name and description. Users could always run Get-VSTeamProcess | FL * to get the current list view.
> 
> This is really nice work.

No problem, I'll add that.

fifthstreetpartners:
Ready for review

DarqueWarrior:
Make sure you import the module and run Invoke-Pester from the Unit\Test folder. I see two tests failing.

DarqueWarrior:
The tests were just missing a mock. I put them in there. Don't worry about it.

fifthstreetpartners:
Sorry I missed that test.  Anything else you want me to change?

DarqueWarrior:
Merging into master now. Thanks so much for your support.

PR #144, opened 2019-02-18
Add security related commands

MichelZ:
# PR Summary

Adding Security related functions

## PR Checklist

- [X] [Update VSTeam.psd1](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#add-a-format-file)
- [X] [Update CHANGELOG.md](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#add-a-format-file)


MichelZ:
@DarqueWarrior Work in Progress, but I thought I'd share this already so you can give initial feedback. Thanks!

DarqueWarrior:
I put in a check that validates the help files are created with a title that matches the filename. This is to catch copy and paste errors. But it also catches empty files. So if you want the build to succeed you are going to have to fill in the help files.

What is up with Get-VSTeamUser2? Not sure why you are not just changing the current function.

MichelZ:
Yes, I will fill them ASAP.

I wasn't sure about changing Get-VSTeamUser, because it does things a little different than Get-VSTeamUser2.
How are your feelings about breaking changes? e.g. rename Get-VSTeamUser to Get-VSTeamUserEntitlement, or Get-VSTeamUserLicense or something like this? That seems to resemble more what the function currently does, and then rename Get-VSTeamUser2 to Get-VSTeamUser ?

DarqueWarrior:
Breaking changes are ok as long as the version number reflects it and it is clearly described in change log.

DarqueWarrior:
We also need to be very clear in the docs the difference between what is being returned from each function. On the surface the results look very similar.

DarqueWarrior:
How is this work coming? Seems like a huge change.

MichelZ:
Still working on it. Got a couple more things that I'd like to add.
If you think it's too large, let me know and we can try splitting it up a bit

MichelZ:
@DarqueWarrior I think this one is ready for a deeper review now

DarqueWarrior:
Sorry for delay I hope to have time this weekend.

MichelZ:
Any luck in getting this one moving?

DarqueWarrior:
Just downloaded to my laptop to review on my flight. Sorry for the delay.

DarqueWarrior:
Can you rename users2.tests.ps1?
APIs not supported on TFS or AzD Server need to throw the error below. 
throw 'This account does not support xxx.' 

You can review Get-VSTeamUserEntitlement for an example of how to throw the corret error.  Now Get-VSTeamSecurityNamespace for example just throws a 404 which is unclear if the API is not surpported or we created the wrong URI for the call.

MichelZ:
Hi Donovan

Are you sure the Get-VSTeamSecurityNamespace is not supported on TFS?
https://docs.microsoft.com/en-us/rest/api/azure/devops/security/security%20namespaces/query?view=vsts-rest-tfs-4.1

This seems to suggest otherwise?

DarqueWarrior:
I will test again. I can give you access to my TFS 2017 and 2018 instances so you can test as well.

MichelZ:
That would be very nice of you and extremely helpful

MichelZ:
@DarqueWarrior Can you get me these login credentials? Thx!

MichelZ:
@DarqueWarrior OK, the issue is that TFS can only query security namespaces by ID, whereas AzD can actually <LIST> them. I think querying them by ID has not much value, so I made that whole command "AzD-only".
Also tested the commands with TFS and fixed a bug encountered while using TFS.

DarqueWarrior:
I see there is a warnning. Source\Public\Remove-VSTeamAccessControlList.ps1(1,10): Warning : Function 'Remove-VSTeamAccessControlList' has verb that could change system state. Therefore, the function has to support 'ShouldProcess'.

MichelZ:
Fixed. I'm not sure why it throws there now though...
```

` Context _handleException
      [-] Should Write two warnings 11ms
        NullReferenceException: Object reference not set to an instance of an object.
        MethodInvocationException: Exception calling "GetParamBlock" with "1" argument(s): "Object reference not set to an instance of an object."
        at Mock, /home/vsts/.local/share/powershell/Modules/Pester/4.7.3/Functions/Mock.ps1: line 237
        at <ScriptBlock>, /home/vsts/work/1/s/unit/test/common.Tests.ps1: line 89
```

    Context _handleException message only
      [-] Should Write one warnings 10ms
        NullReferenceException: Object reference not set to an instance of an object.
        MethodInvocationException: Exception calling "GetParamBlock" with "1" argument(s): "Object reference not set to an instance of an object."
        at Mock, /home/vsts/.local/share/powershell/Modules/Pester/4.7.3/Functions/Mock.ps1: line 237
        at <ScriptBlock>, /home/vsts/work/1/s/unit/test/common.Tests.ps1: line 104`

DarqueWarrior:
I will look into it.

DarqueWarrior:
When unit tests only fail on Mac or Linux it can sometimes be a casing issue.

DarqueWarrior:
Strange I just ran on my local Mac and Ubuntu machines and they all pass.

DarqueWarrior:
There is something going on with the latest version of PowerShell core. If you install 6.2.0 you will get the same errors. Not sure why this is broken. Maybe @SteveL-MSFT knows if something changed in 6.2.0 that could cause this error.

DarqueWarrior:
I think I have a way to test the code without mocking ConvertFrom-JSON which is causing the Null reference exception. Something still seems to be off or different in version 6.2.0 of PowerShell core but I don't want to hold this PR until that is investigated. I will work on new tests this weekend.

SteveL-MSFT:
@DarqueWarrior can you summarize the issue you're seeing with PSCore6.2?

PR #156, opened 2019-05-03
Condtionally set property in VSTeamBuildDefinitionProcessPhase

danstur:
VSTeamBuildDefinitionProcessPhase

# PR Summary

Fix for an issue where the `condition` property is not being set under Azure DevOps server under some circumstances (old builds, or possibly something to do with the update from TFS2018 to AzD, didn't investigate further after hitting the error). This causes an error when creating a `VSTeamBuildDefinitionProcessPhase` object. This PR makes the assignment of the variable conditional, which avoids the error.

## PR Checklist

Not sure if a simple bug fix warrants mentioning in the Changelog, there's certainly no need to update the help and the psd1 also keeps unchanged. The test data was adapted


DarqueWarrior:
Reviewing now.

Issue #185, opened 2019-08-16
[Improvement] Improve Pipelines with CredScan


SebastianSchuetze:
Hi,

I  am currently testing the Microsoft Security Code Analysis, which is in private preview with Microsoft. I ran it on your repository and found some issues with CredScan.
This one tries to find passwords in your repository. 

This one has found 8 issues and I am only posting it here because I checked them all and those where only no real passwords and mostly for the unit tests:

![image](https://user-images.githubusercontent.com/7256168/63196490-b1e82b00-c075-11e9-834c-bb3d3f4e2b5d.png)

Nevertheless, I would like to suggest to implement CredScan to you yaml pipeline and also add a ignore file to it.

I already created it but would like to first align with you, because you would need to install it on you AzD instance first. It is not publically on the marketplace yet, but I am sure it would be no problem for you to use it. I am sure it could be valuable.

SebastianSchuetze:
I added the pull request to show what I meant. I hope that this was okay. Of course, the PR fails, since these tasks are not installed on your organization.

DarqueWarrior:
Do you know who to contact to have the extension shared with the LoECDA org?

SebastianSchuetze:
Yes, initially I signed up at the following page: https://secdevtools.azurewebsites.net/
And the mail given on that page comes when clicking "Sign Up". Never had a direct name, the signature had only "Microsoft SCA Team".

DarqueWarrior:
Closed with pr #186

PR #186, opened 2019-08-16
Topic/add cred scan to build

SebastianSchuetze:
# PR Summary

Including Credscan from the Microsoft Security Code Analysis from Microsoft, which enables to check for credentials in source code, thus improving code quality.

Related to: #185 

## PR Checklist

since this is only improving the pipeline, the checklist would not apply.


DarqueWarrior:
I will review ASAP

DarqueWarrior:
I think I found the people to Email. But if you know for sure who to contact let me know so I can get the extension added to the org.

SebastianSchuetze:
@DarqueWarrior I added also a documentation page to the new CredScan task if people need to add suppression rules for unit tests. 
Please tell me if I put it in the correct folder (I created the folder "build").

Also, I do not fully understand the suppression rules for "hash" type suppressions. If you want to use file-based suppression then we should think of a constants file for passwords that can be used within unit tests. Otherwise, we need to add many files to the suppression rules in the future.

What would be your suggestion?

And thanks for the quick installation of the task! 😄

DarqueWarrior:
I like the idea of a constants file with fake creds to use in tests.

Issue #196, opened 2019-09-09
[Bug] Update-VSTeamWorkItem and Add-VSTeamWorkItem not working properly


SebastianSchuetze:
I caused the problem myself when I committed PR #190 , because I didn't cover the fact when only the parameter AdditionalFields is provided, that in the [code ](https://github.com/DarqueWarrior/vsteam/blob/6448cb0af13e1ab15d3a8781cdeca16899c8108f/Source/Public/Add-VSTeamWorkItem.ps1#L99) $body is expected to be an array. But if I only give one of the fixed fields for the work item (e.g. Description here), then the lines [55-75](https://github.com/DarqueWarrior/vsteam/blob/6448cb0af13e1ab15d3a8781cdeca16899c8108f/Source/Public/Add-VSTeamWorkItem.ps1#L55) create a hashtable instead of an array. So the type has changed.

We should make sure that the outcome is always an array. The problem applies for add and update VSTeamWorkItem. As well if you use only "ParentId" parameter instead of "AdditionalFields".

I will take care of the bug and improve the unit tests to cover this.

@DarqueWarrior is there a way to see which lines of codes are not covered by the unit tests? Then it would be easier to be sure what has been tested.

Steps to reproduce
------------------

```powershell
$additionalFields = @{}
$workItemId = "SomeId"
Update-VSTeamWorkItem -Id $workItemId -Description "Some text" -AdditionalFields $additionalFields
```

Expected behavior
-----------------
Work Item should be updated

Actual behavior
---------------

An error appears saying
```none
Item has already been added. Key in dictionary: 'path'  Key being added: 'path'
At  C:\Users\User\Documents\WindowsPowerShell\Modules\VSTeam\6.3.3\vsteam.functions.ps1:7397 char:16
+                $body += @{
+                ~~~~~~~~~~~
+ CategoryInfo          : OperationStopped: (:) [], ArgumentException
+ FullyQualifiedErrorId : System.ArgumentException
```

Environment data
----------------

OS

- [ ] macOS
- [X] Windows
- [ ] Linux

Server

- [ ] TFS 2017
- [ ] TFS 2018
- [ ] Azure DevOps Server
- [X] Azure DevOps Service

```powershell
> Get-VSTeamAPIVersion
Name                           Value
----                           -----
Release                        5.1-preview
Version                        VSTS
MemberEntitlementManagement    5.1-preview
DistributedTask                5.0-preview
Core                           5.0
Packaging                      5.1-preview
[...]

PR #197, opened 2019-09-09
bug fix for update and add workitem

SebastianSchuetze:
Is connected to the issue #196 which fixes the described problem

### DIFF
--- a/Source/Public/Add-VSTeamWorkItem.ps1
+++ b/Source/Public/Add-VSTeamWorkItem.ps1
@@ -91,7 +91,7 @@ function Add-VSTeamWorkItem {
          foreach ($fieldName in $AdditionalFields.Keys) {
 
             #check that main properties are not added into the additional fields hashtable
-            $foundFields = $body | Where-Object { $_.path -like "*$fieldName" }
+            $foundFields = $body | Where-Object { $null -ne $_ -and $_.path -like "*$fieldName" }
             if ($null -ne $foundFields) {
                throw "Found duplicate field '$fieldName' in parameter AdditionalFields, which is already a parameter. Please remove it."
             }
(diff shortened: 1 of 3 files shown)


SebastianSchuetze:
@DarqueWarrior the build fails for the unit tests. The error is so strange to me, that I don't know what the problem is:
https://loecda.visualstudio.com/Team%20Module/_build/results?buildId=1636&view=logs&j=39439fd4-3ef0-56a3-10f8-5cf61d2586e7&t=86c182c9-17cb-535d-0483-c36562050943&l=1618

I know what it says, but with the unit tests provided it should not be a problem. In theory, they should be correct! Do you have any idea?

DarqueWarrior:
I will find time to review today.

DarqueWarrior:
I get the same errors when I run the test locally as well. Are they passing for you on your machine?

DarqueWarrior:
The tests are failing correctly. In the Assert-MockCalled you are testing for 
```
$Body -like '*Testing*' -and
```

The body does not contain **Testing** so that Assert will fail. On line 28 un-comment out that line so it will show you the REST Call and you can see what you can and can't test for. 

Just make sure and comment it out again before submit.

DarqueWarrior:
You would have to assert each part of the request separate and fail just that part. The quick way around this is to uncomment the line in the mock until the test is working.

SebastianSchuetze:
@DarqueWarrior unit tests pass now, but the Package Management Stage seems to fail.
Btw. how does it work with GitHub? Shouldn't only the build stages be tested on the Azure Pipelines instead of the whole pipeline?

DarqueWarrior:
Yeah I need to only have the build run. The release is not going to work for permission reasons. I will work to fix that.

DarqueWarrior:
Please make sure I permission to push to this PR.

DarqueWarrior:
Not sure why I was getting an error? hmm.

DarqueWarrior:
remote: Permission to RazorSPoint/vsteam.git denied to DarqueWarrior.
fatal: unable to access 'https://github.com/RazorSPoint/vsteam.git/': The requested URL returned error: 403

SebastianSchuetze:
What changed? Where there some permission errors?

PR #204, opened 2019-10-04
add maxParallelism to Disable/Enable-VSTeamAgent

### LINKED ISSUE #203: Disable/Enable-VSTeamAgent results in unusable agent due to maxParallelism reset to 0
Steps to reproduce
------------------

```powershell
Disable-VSTeamAgent -poolId xx -Id yy
Enable-VSTeamAgent -poolId xx -Id yy
```
Queue a build to this specific agent

Expected behavior
-----------------
- the agent to start building

Actual behavior
---------------
- the agent seems to be "stuck" and is not picking up any build anymore
- reboot does not help
- disable/enable through web interface does not help
- reinstalling agent does "fix" it

After searching for a while we discovered that it seems the `maxParallelism` is reset to `0` while using Disable/Enable-VSTeamAgent. 

As described on https://docs.microsoft.com/en-us/rest/api/azure/devops/distributedtask/agents/update?view=azure-devops-rest-5.1#request-body `maxParallelism` is "Maximum job parallelism allowed for this agent.".

I don't know why it resets to `0`, but adding 'maxParallelism':1 to Disable/Enable-VSTeamAgent.ps1 did solve the issue, but i am wondering if anyone else is seeing this same behaviour?

Environment data
----------------

OS
- Windows

Server

- Azure DevOps Server RTM

```powershell
> Get-VSTeamAPIVersion
Release                        3.0-preview
Version                        TFS2017
MemberEntitlementManagement
DistributedTask                3.0-preview
Core                           3.0
Packaging
ServiceFabricEndpoint
VariableGroups                 3.2-preview.1
Build                          3.0
Git                            3.0
ExtensionsManagement           3.0-preview
Graph
Tfvc                           3.0

```

```powershell
> $PSVersionTable
PSVersion                      5.1.17134.858
BuildVersion                   10.0.17134.858
```

@Geertvdc have you perhaps also seen this behavior on an Azure DevOps Server environment?

japj:
# PR Summary

Fixes #203 by adding `maxParallelism` to Disable/Enable-VSTeamAgent

What would be the best way to extend any existing tests for this?


DarqueWarrior:
Is this a value that the caller should be able to set? We can have it default to 1 but is there a use case where I might want to set it?

japj:
I don’t know, from the web ui I have only seen a value 1 yet.
It could be that we need to fetch current settings of the agent, patch in the enabled flag and send that back to keep the current setting?

The weird thing is that without this maxparralism gets reset to 0 (which might actually be a server bug?).

dickvdm:
What are the properties of an agent that are settable by the caller, which are read-only? What does maxParallelism mean anyway ;-)? @bryanmacfarlane, is there documentation on this? 

For XAML agents the **offline/online** was also caller settable, which we loved and used heavily in following scenario:
-Set all agents on a specific host to _offline_ status.
-Do host updates (Operating system, application, configuration, etc.)
-Reboot the host. After reboot the agents on this host would come back to _online_ status again, ready to accept jobs on a freshly updated host :-).

On microsoft/azure-pipelines-agent only **enable/disable** of agents is caller-settable, **offline/online** is not afaik. So we have to align _host reboot_ and _agent enabling_ ourselves. As we can't do this in an atomic action, the annoying scenario can occur that scheduled/running builds fail because the host is rebooting.

DarqueWarrior:
Sorry for delay. I am recovering from surgery. I will get to this ASAP.

PR #209, opened 2019-10-19
Added CMDLET to delete workitem

SebastianSchuetze:
# PR Summary

I added a new cmdlet to delete work items by IDs. I

So the following things were added:
* cmdlet Remove-VSTeamWiql
   * added also the switch parameter "Destroy" to also permanently delete the work item instead on putting it to the recycle bin
* added a format file for the cmdelet
* added a type file for the cmdlet

## PR Checklist

- [X] [Update VSTeam.psd1](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#add-a-format-file)
- [X] [Update CHANGELOG.md](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#add-a-format-file)


SebastianSchuetze:
@DarqueWarrior I accidentally merged the other topic branch into this one in the beginning. So it shows the history of the not yet merged PR and also the comments above. I don't know if that will be a problem.

DarqueWarrior:
Make sure you run

```PowerShell
.\Build-Module.ps1 -analyzeScript -ipmo
```
And clear any messages. 

I am a big unclear why we need _applyTypesToWorkItemDeleted and can't use _applyTypesToWorkItem

DarqueWarrior:
I will take care of these changes.

SebastianSchuetze:
Just for the explanation. I created _applyTypesToWorkItemDeleted because the return JSON object is not consistent with a normal work item object. Strange enough, even the property names are different.
for example, the name of the work item when you query it is "title". But the object returned from the API when you delete it has the property named "name" and not "title"

This is for several properties. So I thought a new method is needed because I also created a new format.

Unless you tell me, that this can be done differently, then I am happy to know how it can be done simpler.

DarqueWarrior:
Can you think of a scenario where you are going to save the returned deleted item? In all the rest of the remove calls we don't return the object. I think to stay consistent I will just not return anything here either.  I am almost done with all the changes. There is nothing for you to do at this point.

DarqueWarrior:
Changes merged into PR #217

Issue #148, opened 2019-03-08
Get-VSTeamBuildArtifact fails


apobekiaris:
Hi

I am executing the next script 

```ps1
$build = (Get-VSTeamBuild -ProjectName eXpandFramework|Where-Object {$_.DefinitionName -like "Xpand-Lab" -and $_.Result -eq "succeeded"}|Select-Object -First 1)
Get-VSTeamBuildArtifact -ProjectName eXpandFramework -Id $build.id 
```

however there is an exception in the output
```
The property 'properties' cannot be found on this object. Verify that the property exists.
At C:\Program Files\WindowsPowerShell\Modules\VSTeam\6.1.0\Private\applyTypes.ps1:93 char:7
+       $item.resource.properties.PSObject.TypeNames.Insert(0, 'Team.Bu ...
+       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], PropertyNotFoundException
    + FullyQualifiedErrorId : PropertyNotFoundStrict


id   : 232
name : Xpand.v18.2.601.10


id   : 233
name : build.SourceLabel
```

and is still not clear how do I download the artifacts.

DarqueWarrior:
I think this has to do with upgrading the version of the API we use by default. I have noticed some properties have moved around. I will look into it now.

DarqueWarrior:
Please run the following command and report the results.

```PowerShell
Get-VSTeamAPIVersion
```

apobekiaris:
```
Name                           Value
----                           -----
ServiceFabricEndpoint
Build                          3.0
ExtensionsManagement           3.0-preview
Release                        3.0-preview
Version                        TFS2017
MemberEntitlementManagement
Packaging
Tfvc                           3.0
Git                            3.0
Core                           3.0
Graph
DistributedTask                3.0-preview
```

apobekiaris:
@DarqueWarrior any updates on this case?

looks like if I ErrorAction Continue I get a valid download Url

DarqueWarrior:
Are you connected to a TFS2017 server?

apobekiaris:
[...]

PR #232, opened 2020-01-23
Fix for #148

markwragg:
This fixes a bug in `Get-VSTeamBuildArtifact` where an error is returned because the API returns an extra record along with the list of artifacts that is called 'build.sourcelabel' and contains a URL but no "properties" object.

This PR stops the error by excluding any build artifact records returned by the API that don't contain a "properties" object from getting the build artifact specific type names applied. The result is still returned to the user however.

Alternatively if we think it would be better to exclude this result entirely, then it could instead be filtered out in the `Get-VSTeamBuildArtifact` function.

# PR Summary

## PR Checklist

- [ ] [Update CHANGELOG.md](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#add-a-format-file)

### DIFF
--- a/Source/Private/applyTypes.ps1
+++ b/Source/Private/applyTypes.ps1
@@ -95,7 +95,7 @@ function _applyTypesToBuild {
 function _applyArtifactTypes {
    $item.PSObject.TypeNames.Insert(0, "Team.Build.Artifact")
 
-   if ($item.PSObject.Properties.Match('resource').count -gt 0 -and $null -ne $item.resource) {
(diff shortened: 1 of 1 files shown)


SebastianSchuetze:
HEy @markwragg . I am not the maintainer, but maybe it could be good to add a unit test to test exactly for the problem you to fixed? Would tat be possible?

SebastianSchuetze:
@markwragg would you able to add a unit test for this? Do you need help? I am also now a maintainer of this repo. So I could help you much better with problems regarding your PR.

MichelZ:
I've added a test for this in my PR #257 
Feel free to copy it into your PR and to merge this one :)

PR #265, opened 2020-02-25
Simplify merging of files

MichelZ:
# PR Summary

Simplify merging of files


Source/Classes/_classes.json
```diff
@@ -2,62 +2,8 @@
    "outputFile": "vsteam.classes.ps1",
    "fileType": "classes",
    "files": [
-      "VSTeamVersions.ps1",
-      "VSTeamProjectCache.ps1",
-      "VSTeamProcessCache.ps1",
       "VSTeamDirectory.ps1",
```
MichelZ: Just FYI - It is intentional to leave these two in, as they are base classes for the others and powershell won't work properly if they are "out of order".
The file list now gets filtered for unique values first.

DarqueWarrior:
The reason the _classes.json exist is because the classes had to be loaded in a particular order. If you notice the files are not loaded in the order they would be listed in a *.ps1 search.

DarqueWarrior:
> @DarqueWarrior Yes, I know.
> This is why I added a feature to be able to list the base classes explicitly, but load everything else using *.ps1. As you can see, the unit tests still work :)

The newer unit tests load the needed classes explicitly. 

```PowerShell
$here = Split-Path -Parent $MyInvocation.MyCommand.Path
$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path).Replace(".Tests.", ".")

. "$here/../../Source/Classes/VSTeamVersions.ps1"
. "$here/../../Source/Classes/VSTeamProjectCache.ps1"
. "$here/../../Source/Private/common.ps1"
. "$here/../../Source/Public/$sut"
```

Those tests are not testing the module loading. The classes that are needed are loaded in the correct order for the tests. If those classes are loaded in the wrong order everything will fail.

DarqueWarrior:
It looks like the code can be used for some files just not the classes.

MichelZ:
@DarqueWarrior hmmm...   ok, I'll have a closer look then and eventually back out the classes.
Thanks for the hints

MichelZ:
@DarqueWarrior I've removed the classes changes

PR #266, opened 2020-02-26
Improve setup/teardown, add -runIntegrationTests to Build script

MichelZ:
# PR Summary

- Improve setup/teardown, add -runIntegrationTests to Build script
- Be less destructive, only destroy projects that begin with `TeamModuleIntegration`

@DarqueWarrior Let's try this again :)
This time I was able to run the tests.
Any way you can run the integration tests on this PR manually before merging?


MichelZ:
@SebastianSchuetze  @DarqueWarrior

SebastianSchuetze:
we see you, but we do this on the side. So please wait. Since I am pretty new as maintainer I am dependent a bit on Donovan's opinion until I know how to handle certain code changes. This one actually caused a problem before.

SebastianSchuetze:
I have seen your last commit trying to make things less destructive. It does not make sense here. For a full setup and destroy philosophy of a test environment you would even create a new Azure DevOps tenant. If not then expect that nothing stands anymore after the integration test.
Could you revert these changes?

There are multiple changes in that PR which make it hard for me to understand what happens there. 

Also, would you be so kind as to solve the conflicts? I will try afterwards to get an opinion from Donovan.

SebastianSchuetze:
@MichelZ I included the change for the integration script into the build script but changed the approach a little bit. Also due to some changes that happened between your commit and today's state. I also did not include your integration test changes since @DarqueWarrior changed some things that are similar to yours.
Also, I removed that flag from running the unit tests automatically because if you want to run both you can set the runTest parameter added to the integration test parameter.

@DarqueWarrior I think the addition into the build script is nice if you want to run the integration tests locally on your machine

PR #269, opened 2020-02-26
Retrieve stale branches

MichelZ:
# PR Summary

A feature to retrieve branches for projects/repos that have not been committed to in more than X days

## PR Checklist

- [x] [Update CHANGELOG.md](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#add-a-format-file)


Source/Public/Get-VSTeamStaleBranch.ps1
```diff
@@ -0,0 +1,89 @@
...
+                        LastCommitDate = $branchStats.commit.committer.date
+                        Ahead = $branchStats.aheadCount
+                        Behind = $branchStats.behindCount
+                     }
+
+                  _applyTypes $object "Team.GitStaleBranch"
+                  Write-Output $object
+               }
+            }
+         } elseif ($ProjectName) { # Retrieve whole project (recursive)
```
SebastianSchuetze: @MichelZ it would be better if we use cmdlets that are more atomic and simpler. 
Could you please change it so that always a repository ID is needed and a project is used?

When this is changed that way I would let the scripting person choose to add the logic himself.

unit/test/Get-VSTeamStaleBranch.Tests.ps1
```diff
@@ -0,0 +1,168 @@
...
+      Repository = "repo"
+      LastCommit = "Block-SmbShareAccess"
+      LastCommitter = [PSCustomObject]@{
+         displayName = 'Microsoft.VisualStudio.Services.TFS'
+         id          = '1'
+         uniqueName  = 'some@email.com'
+      }
+   }
+
+   $singleRepository = [VSTeamGitRepository]::new([PSCustomObject]@{
```
SebastianSchuetze: When you have complex objects for tests. You might want to save an example JSON in the subfolder (examplefiles](https://github.com/DarqueWarrior/vsteam/tree/master/unit/test/sampleFiles). This makes the unit test simpler to understand and also reusable if other unit tests need to use it.

SebastianSchuetze:
@DarqueWarrior just to be sure. How do you see my reasoning? I had the feeling that pretty much most of the cmdlets are much simpler.

Source/Public/Get-VSTeamStaleBranch.ps1
MichelZ: But why though? This is optional, and you're free to use the repoID if you want to...

unit/test/Get-VSTeamStaleBranch.Tests.ps1
MichelZ: True, will see that I can do that in my next revision

DarqueWarrior:
> @DarqueWarrior just to be sure. How do you see my reasoning? I had the feeling that pretty much most of the cmdlets are much simpler.

Currently we are just "wrapping" the REST API of Azure DevOps. However, I am not opposed to pushing this module beyond that. This might make even more popular. Let me review this PR and see how I feel.

But to answer your question I am not opposed to "super-luxury" functions as you call them. :)

DarqueWarrior:
I also think if we rename and distance ourselves a bit from the product and position the module more around DevOps the practice... Then functions that do more makes even more sense. 

What harm is there to provide functions that achieve a particular goal that might be a combination of calls?

SebastianSchuetze:
I was more worried this might make the code more complicated and feared that it might reduce performance. But it is more a fuzzy feeling because I like to wrap recurring tasks as well in well-written functions.
So I like the idea if we can hold quality still high.

MichelZ:
If there‘s something wrong with my code quality? 🧐

SebastianSchuetze:
Haha, not necessarily yours @MichelZ. When we open up this approach, then more complex cmdlets will be created by more people. It's got for added value when added complexity does not reduce code quality. That is all.
But I trust also when Donavan and other contributors like you believe it is manageable.

DarqueWarrior:
```PowerShell
PS C:\> Get-VSTeamStaleBranch -top 5 | Format-Wide
```

This will return the top five Process Templates only showing their name

That is not what this example does. Looks like a copy and paste error. Also there is no -top parameter on this function.

DarqueWarrior:
Do you want list vs table to be the default output of the function?  I think the majority of the functions default to table.

DarqueWarrior:
Now I remember why formats were listed explicitly. Using the "*.ps1xml" they are combined in alphabetical order which means List will be first in the file and that is the default view now.

DarqueWarrior:
I am releasing a 6.4.6 today to correct the List vs Table bug.

smurawski:
@DarqueWarrior, @MichelZ, @SebastianSchuetze,

I like the function, but I think it (and other functions like it) belong in a different module.

This module is effectively the PowerShell SDK for the Azure DevOps APIs.  The higher level, task-based functions should be in a separate module that depends on a particular version of the VSTeam module.  This allows the two efforts to evolve at different paces.

There are two main problems I have with bringing the higher level, task-based functions into VSTeam:

- Tight coupling/expanded cross-module dependency - every iteration needs to ensure the behavior of those higher level functions.
- Adding new higher level functions adds cognitive load to everyone using the module.  Every get-command -module vsteam or tab complete from get-vsteam.... increase what you have to filter through to find things.  Starting a new module makes it more likely that command discovery for high level tasks is easy - even as the scope grows.

Just my thoughts on the matter - I think the capability is great, but I don't think this module is where it should live.

SebastianSchuetze:
Thanks for the insight @smurawski. This is somehow what I was worried about. So you also think that it would get hard to maintain if such functions are added?

How would that work together, when two modules depend on each other? I see that working with e.g. the Az modules. But do you have any clue about the inner workings of this?

DarqueWarrior:
Module dependency is handled in the PSD1 file. VSTeam depends on SHiPS for example. The other module would just list VSTeam as a dependency and when a user installs one the other gets installed first.

smurawski:
@SebastianSchuetze PowerShell module dependencies are kind of a bad experience due to version specificity constraints, but it is better than nothing.  As Donovan notes, you just specify the dependency in the module metadata.  That will help when you install the module from the powershell gallery, ensuring that you have the dependencies locally or pulling them down as well as making sure it is present and loaded at run time.

SebastianSchuetze:
@DarqueWarrior since we had the discussion internally and also here and we came to the conclusion that this module is currently only a wrapper. Maybe this function could belong to the set of example scripts. Please decide if this PR should be closed.

DarqueWarrior:
I am creating a repo called vsteam-plus where we can put functions like this one.

SebastianSchuetze:
@MichelZ I am closing this PR, but please keep it for now for you. Soon we will have another module building up on this one that can support epic shit like yours. ;-)
I also marked it with a new tag called "vsteam plus" to find all features that should go in this module in the future @DarqueWarrior

DarqueWarrior:
I am getting the pipeline for vsteam-plus setup now. But I think GitHub is having an LSI. My last few commits are not showing up in GitHub...

PR #253, opened 2020-02-24
Add GitHub release

MichelZ:
# PR Summary

Add the release to GitHub as well.
Fixes #251 

Note: You need to create a GitHub service connection named "vsteam" for this to work before merging/releasing the PR!!

## PR Checklist

- [ ] [Write Help](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#write-help)
- [ ] [Write Unit Test](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#write-unit-test)
- [ ] [Update CHANGELOG.md](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#add-a-format-file)

(not needed)

PR #270, opened 2020-02-26
Pass variables cross stage

tomkerkhove:
# PR Summary

Pass variables across stages in Azure DevOps pipeline so tat `PACKAGE_VERSION` is available in other stages as well.

This is required to fix the current GitHub releases.

Follow-up for #253 to fix broken GitHub releases.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -188,6 +188,13 @@ stages:
               publishVstsFeed: 'vsteam'
               allowPackageConflicts: true
 
+          - template: ./templates/persist-variable.yml
+            parameters:
+              variableName: 'PACKAGE_VERSION'
+
+          - publish: $(Pipeline.Workspace)/variables
+            artifact: variables
+
 - stage: Testing
   displayName: Testing stage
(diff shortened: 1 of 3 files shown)


SebastianSchuetze:
@tomkerkhove nice solution! Somehow I feel it's a lot just to get the version number going. Wouldn't it be easier to e.g. get the version number always directly from the psd1 in the PowerShell package?

tomkerkhove:
It's up to you guys but I think it's always interesting to do it as part of the start and re-use where need be, but it's just a suggestion.

DarqueWarrior:
I think we have this solved but adding 

```Powershell
Write-Host "##vso[task.setvariable variable=PACKAGE_VERSION;isOutput=false]$($manifest.ModuleVersion)"
```
To the PowerShell task in the Production stage. After I push 6.4.5 if there is still a problem I will return to this PR. If it works I will simply close this.

Either way thanks for the support.

DarqueWarrior:
This has been fixed. Thanks

Issue #262, opened 2020-02-25
Code Coverage


MichelZ:
Is there a reason that Code Coverage is disabled?
(`e.g. build/unitTest.yml -> parameters: CodeCoverage: 'false'`)

if not - can this be enabled?

SebastianSchuetze:
Check the following issue:
https://github.com/pester/Pester/issues/1318
and
https://github.com/PowerShell/PowerShell/issues/9884

Since an OS update for windows PS is getting very flow when loading classes. Since the script analyzer is doing that very often, the current build can take very long time (30-50 minutes) because of that.

I think we are dependent from this issue being solved.

MichelZ:
@SebastianSchuetze This has been fixed in `PS Core 6.2.3` according to your link(s), `6.2.4` is currently installed on the [Windows 2019](https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md) and [Windows 2016](https://github.com/actions/virtual-environments/blob/master/images/win/Windows2016-Readme.md) images. 

I think we should be good here?

SebastianSchuetze:
I hope that this was also the intention of @DarqueWarrior because of the deactivation. Let's wait for his opinion to be sure that there is nor another reason.

DarqueWarrior:
I can do some testing and see if that has been fixed. The builds were taking way way too long.

MichelZ:
@DarqueWarrior Success?

SebastianSchuetze:
Currently the limitation of code coverage is a limitation due to Pester and PowerShell. So this cannot be improved currently without too much effort (parallelization of unit tests with agents and merging results).

We will see if we find something to improve in the future.

PR #282, opened 2020-03-13
Activated code coverage

SebastianSchuetze:
# PR Summary

This is to check whether Pester is still slow with code coverage.

#262 is related.

## PR Checklist


SebastianSchuetze:
I made a test and basically switched the PowerShell task to use PowerShell core. 

You can see the difference when using PS 5 on windows takes roughly 24 minutes (whole pipeline 28 minutes)
https://loecda.visualstudio.com/Team%20Module/_build/results?buildId=1872&view=logs&j=df7cdcec-8c9a-591d-7383-990f35cfc117&t=1ad5f601-14eb-5211-34b7-3146063ec0a0

the same with PS core takes only around 10 minutes (whole pipeline 16 minutes) on the same machine:
https://loecda.visualstudio.com/Team%20Module/_build/results?buildId=1873&view=logs&j=df7cdcec-8c9a-591d-7383-990f35cfc117&t=1ad5f601-14eb-5211-34b7-3146063ec0a0

Do you think it that is fast enough again @DarqueWarrior? How long did it take before the bug?

SebastianSchuetze:
I changed it to only have code coverage on the windows machine.

SebastianSchuetze:
@DarqueWarrior what do you say? Is the time enough to turn it on again?

DarqueWarrior:
It is still really slow. I have been testing locally.

DarqueWarrior:
The windows build already takes the longest. Wonder if it would not make sense to enable on one of the other platforms instead so they can be running at the same time.

SebastianSchuetze:
For pester there is an open issue or the plan to separate test on to multiple processes or agents. Not sure how far they are that would speed up things!

As for running the tests on another agent, we could do it. I will modify this PR and switch it to a Mac OS agent. I made tests in the past and Mac OS seemed always the fastest to me.

SebastianSchuetze:
@DarqueWarrior as you can see I ran the code coverage on mac and Unix in parallel.
They even took a bit longer around 20 minutes. So this is not faster, but I got an idea of how to maybe solve it, but it is a bit of effort. Look at  the following graphic

image

So we would split the unit tests into subfolders within the pipeline (not in the repo) dynamically by a divisor X of the number of unit test files.
Then we run multiple unit tests in parallel, collect the results, merge them and then commit them.

What do you say? Too much effort or manageable?

DarqueWarrior:
How much faster do you think this will make it?

SebastianSchuetze:
How much it speeds up depends on the following parameters:
* how many agents can we run maximum in parallel:
  * one for mac
  * one for Unix
  * 8 left for code coverage testing with unit tests

If we expect that only one build pipeline instance is running at a certain point in time. I would generally expect that there are not more than 2 at the same time with the current workload.

I would suggest dividing the workload by at least 4 parallel agents in a [YAML loop on job level](https://github.com/microsoft/azure-pipelines-yaml/blob/master/design/each-expression.md). But how long each package needs really depends on the number of unit tests per file. I would only simply divide the number of unit test files by X and then put them in subfolders.

We also have some overhead for each job, since we have to install all the needed modules every time (e.g. Pester).

In the best case, it would then be around 3 minutes for all agents and in worst case 5 minutes for all. It would get better when we use more agents.

I could give it a try and prototype something.

DarqueWarrior:
The performance is still too slow.

PR #305, opened 2020-05-07
Topic/scaffold template with plaster

### LINKED ISSUE #292: Suggestion: Create Scaffolding Script for new cmdlets
I suggest using plaster to scaffold all files needed for new cmdlets

https://github.com/PowerShell/Plaster/blob/master/

I suggest the following things to scaffold from templates:
* cmdlet file with name Verb-VSteamAction
* create markdown docs with synopsis file and basic doc file
* check whether during scaffold we can already define parameters for the cmdlet and also add it to the files (cmdlet parameters, add them to the documentation already)
* add a prepared unit test file
* ask for if a types/formats file should be created

With plaster, we would define a manifest that defines the scaffolding. So we would need the following things:

1. plaster manifest file
2. cmdlet template file
3. docs template files
4. unit test template file
5. type XML template file
6. formats XML template files

What do you think @DarqueWarrior. This could make it easier for people to get started and we are more standardized.

SebastianSchuetze:
# PR Summary

Adding Plaster for scaffolding new commands for the module. Also adjusted the documentation for the contribution guide to making this mandatory for every new function.

In this case we make sure:
* function template file is used
* unit test file template is used
* documentation template is generated
* optionally - types and formats for returned data is generated

It closes #292 

## PR Checklist

- [ ] [Write Unit Test] - not needed
- [ ] [Update CHANGELOG.md] - not needed


SebastianSchuetze:
@DarqueWarrior can you check the template if you think it applies to your desired coding standards?

DarqueWarrior:
First this is awesome!

Some things I noticed.

When I selected Get as the cmdlet the sample code had
```
-Method "Post"
```

Should the selection of cmdlet be a freeform? I know there is an "approved" list of verbs but you can use "non-approved" verbs as well. I also noticed 'Invoke' was not listed and we have one of those. 

What are your thoughts on adding Plaster as a dependency so it is downloaded for you when you install VSTeam?


Again this is a great addition.

SebastianSchuetze:
Yes, we could add Plaster as a dependency. We have 2 options with Plaster that I can see:
* Use templates for new developments for the module
* Use templates included in the module which gives you jumpstart scripts when creating your own scripts based on VSTeam

Having said that to use Plaster as a dependency where we only use templates not included in the module does not make much sense (we also don't have Pester as a dependency). Unless we have jumpstart scripts for module users as well.

Yes, I wanted to ensure somehow approved verbs, but if you think it is too strict, then I can just use a free form.

We can go as complex as you want. The Method is just hardcoded in the example.

DarqueWarrior:
We might want to press pause on this until the Pester5 changes are done. The test structure is changing.

SebastianSchuetze:
yes makes sense. I will be waiting to then adjust the unit test template.

SebastianSchuetze:
@DarqueWarrior shall I resume and update the templates? Which unit test file do you think fits to convert it to a proper Pester 5 unit test template?

DarqueWarrior:
I think this code should be its own project in the new org.

SebastianSchuetze:
Ok cool. But we can make another PR once we have a small base of quickstart scripts then we can include them as plaster scripts in the module to have a quickstart for people.

Issue #300, opened 2020-05-02
Improvements to Get-VsTeamProcess and ProcessCache & ProjectCache classes


jhoneill:
It would be useful if Get-VSTeamProcess 

1. Got process information from the /work/processes URI path rather than process/processes as the former has additional details
2. Could query from processes with a wildcard name.
3. Updated the cached process information 
4. Displayed whether a process was disabled or was the default process. 

It would be neater if the cache classes had an `update` method to 
1. call the _getWhatever  mockable function and store the result
2. Update the cache time
(at some point it would also change how the the cache times out. Currently it will time out if the time goes from 15:27:59.99  to 15:28.00.01 but not if it goes to 17:27:30 )

and a `getCurrent` method to decided if the cache is stale, and call update if it is. 

This means instead of having 
```
If (_staleTest) {
  [class]::content = _getSomething
  [class]::timesamp = Something
}
doSomethingWith  [class]::content

```
We can simply use doSomethingWith  `[class]::GetCurrent().`  The old ways still work, for things which need to update or invalidate the cache. 

I have made these changes, and they will also come in my next PR, as with 299, I wanted an issue to link to.

PR #322, opened 2020-05-22
Enhance Get-VSTeamProcess #300

jhoneill:
# PR Summary

Get-VSTeamProcess now:

- Gets information from the /work/processes URI path rather than process/processes as the former has additional details
- Can query for processes using a wildcard name.
- Updates when it runs  the cached process information
- Displays whether a process was disabled or is the default process.

## PR Checklist

- [X] [Write Help]
(Small updates, mostly reflecting the different API provides more information but doesn't support paged responses)
- [X] [Write Unit Test]
(Fixed existing ones.)
- [X] [Update CHANGELOG.md]


.docs/Get-VSTeamProcess.md
```diff
@@ -29,18 +27,32 @@ This will return all the Process Templates
 ### -------------------------- EXAMPLE 2 --------------------------
 
 ```PowerShell
-PS C:\> Get-VSTeamProcess -top 5 | Format-Wide
+PS C:\> Get-VSTeamProcess  | Format-Wide
```
DarqueWarrior, line 30: Why get rid of the -Top 5 example? Just piping to FW is not teaching how to use Get-VSTeamProcess.

Build-Module.ps1
```diff
@@ -100,8 +108,8 @@ if ($ipmo.IsPresent -or $runTests.IsPresent) {
    if ((Get-Module VSTeam)) {
       Remove-Module VSTeam
    }
-
-   Import-Module "$output/VSTeam.psd1" -Force
+   $env:testing = $true
```
DarqueWarrior, line 111: I don't want any testing flags anywhere. The code should be written not to require this. Tests should be able to mock their way around any need for a testing flag.

Source/Classes/VSTeamProcess.ps1
```diff
@@ -4,24 +4,46 @@ class VSTeamProcess {
...
-      $this.Name = $obj.name
-      $this.Type = $obj.type
-
+      $this.AddTypeName('Team.Process')
+      $this.Name                    = $obj.name
+      $this.ProcessTemplate         = $obj.name
+      #Allow for processes from .../process/processes with ID property, or ...work/processes with TypeID
+      if ($obj.PSObject.Properties['typeId']) {
+         $this.ID                      = $obj.typeId
+         $this.URL                     = (_getInstance) + "/_apis/work/processes/" + $obj.typeId
```
DarqueWarrior, line 22: Why are these URLs being hand built and not using the url property of the object returned?

Source/Classes/_classes.json
```diff
@@ -1,5 +1,5 @@
 {
-   "outputFile": "vsteam.classes.ps1",
```
DarqueWarrior, line 2: I am not ok with these changes right now. These change the build process and I am not wanting to change the build process at this time.

Source/Public/Get-VSTeamProcess.ps1
```diff
@@ -1,56 +1,46 @@
...
    [CmdletBinding(DefaultParameterSetName = 'List')]
+   [OutputType([vsteamprocess])]
    param(
-      [Parameter(ParameterSetName = 'List')]
+      [Parameter(ParameterSetName = 'ByName', Position=0)]
+      [ArgumentCompleter([ProcessTemplateCompleter])]
+      [Alias('ProcessName','ProcessTemplate')]
+      $Name = '*',
+
+      [Parameter(DontShow=$true, ParameterSetName = 'List')]
```
DarqueWarrior, line 10: Why are you setting DontShow here? Why would we want to hide this?

Source/Classes/_classes.json
jhoneill: Understood. See my comment on the the other PR about why you might want to keep in mind for another time

Source/Classes/VSTeamProcess.ps1
jhoneill: There are two branches to the URI tree where you can get processes either under work/processes or Process/processes. 
process/processes is read only, uses "ID" for the ID,  includes the URL but doesn't give information about which processes are disabled and which one is the default. 
So I switched it to work/processes which calls the ID "typeID", _doesn't include the URL_ (which I need later when setting thinks like work item types in a process, or process behaviours , or process properties), but has the advantage of returning the same, richer, object which Set- and Add- return.      
If only I knew, say, a PM on that team who could have a word with people and tell them to "stick to one name for IDs and include the xxx URL!" 
For now the only way to put the URL back is ugly.

Build-Module.ps1
jhoneill: That's a hangover which should have been removed.  Either I grabbed the wrong version of build.ps1 after sebastian said he didn't understand why the build was breaking when I left it to you guys to take the "build-classes to PSM1" change or not, or I never took it out. Nothing should be referring to it any more.  
But  from the last few PRs I've put in you want to ignore my build.ps1 , ignore my change to the psm1 file and undo the the change to where _classes.json builds to.  (I'm going to keep two versions, one which works for my development and one to put in PRs for you :-) )

.docs/Get-VSTeamProcess.md
jhoneill: Because the API ignores top. 
When I changed the source for processes (see below for the explanation) I found that -top 5 returned all the processes in a live project.  I also found that $skip=0 never got into the query URI. I know now it is done on purpose ... 
Overnight I had an idea. _If the user specifies top or skip call the process/processes API not the the work/processes API_  Tested that, and I got all the processes. It's not the function, and it's nothing to do with whether skip is there or not. A URI with $top=5 and/or $skip=2 returns everything. 

I can't find the docs page I was looking at this morning which showed top and skip but this one (which I swear is in another part of the tree)  does not mention them 
https://docs.microsoft.com/en-us/rest/api/azure/devops/core/processes/get?view=azure-devops-rest-5.1

So.... the -top and -skip and parameters are now ignored. If someone is using them I don't want their code to break so the parameters are still there but hidden, and they have been expunged from the help .

Source/Public/Get-VSTeamProcess.ps1
jhoneill: See above on the help. 
It looks top and skip are ignored when placed in the URI  (this reference in the docs has airbrushed them out https://docs.microsoft.com/en-us/rest/api/azure/devops/core/Processes/Get?view=azure-devops-rest-5.1 ) 
To avoid breaking any existing scripts the parameters need to remain, but since they will be ignored they're best not shown.

DarqueWarrior: I wonder if when that was written it was a copy and paste error. I reviewed the oldest version of the API we support and can't find those anywhere! https://docs.microsoft.com/en-us/azure/devops/integrate/previous-apis/tfs/processes?view=tfs-2017.  So I will remove those and change the version number to 7.x to denote a breaking change.

Source/Classes/VSTeamProcess.ps1
DarqueWarrior: hmm. I might ping one of the PMs and see what is going one.

jhoneill:
> I do not want to change the build process in any way at this time.

I've changed my build so that 

- _Classes.json is as as you had it and vsteam.classes.ps1 gets built.
- vsteam.psm1 dot sources vsteam.classes.ps1 as you had it 
- vsteam.psm1 is copied as is to the output directory, as you had it.

I have also added a switch to build module **if specified**
- vsteam.psm1 in the output directory is appended to vsteam.classes.ps1
- the now-extended vsteam.classes.ps1 is copied over vsteam.psm1. 
- vsteam.classes.ps1 will still be dot sourced so is replaced with  "#Empty"

So have we can development that little bit easier with a build-time switch and the "normal" build process is unaffected.

> I do not want any testing flags of any kind.

I found one last stray one in Add-VSTeamBuild.ps1  they are all removed from my branches now (I think)

Source/Public/Get-VSTeamProcess.ps1
jhoneill: :-) 

In a spirit of gentle teasing (and **not** criticism because you could find I've done exactly the same) but _"was it ever tested and found to work against a live system"._ 

I have found that rather than removing parameters, changing the version and telling people to change their code, it can better in practice to hide the parameter and write a warning when it anyone uses it saying "This is ignored, and may not be supported in future versions".    
Have you seen how annoying it is in Pester 5 that debug in vscode tries to run `invoke-pester -script blah.ps1 ` and pester says "`-script is an unknown switch`"   but `invoke-pester blah.ps1` works so it would be trivial to keep compatibility with a parameter alias.    
Being too  gung-ho about "break this, break that, but it's OK if I change the version number:... if people don't keep up to date and / or flag a maximum version and make sure they load and old one that's _their_ problem" just annoys the user base. 
Microsoft was based on the premise that old stuff worked with new stuff - CP/M recompiled for DOS, DOS ran under Windows, OS/2 and NT,  OS/2 and Win 3.1 also ran on NT, 32bit Windows runs on 64 bit Windows.  So Microsoft veterans like me (Microsoft user in late 70s, 80s and early 90s, Microsoft certified partner mid and late 90s, Microsoft employee 2000-2010) can get a bit rude if people say backwards compatibility is hard :-)  "You don't know you're born ! In my day....  Nurse! "

DarqueWarrior: The warning idea sounds good. Letting them know these parameters will be removed in future versions.

.docs/Get-VSTeamProcess.md
```diff
@@ -29,18 +27,32 @@ This will return all the Process Templates
...
+
 
 ## PARAMETERS
 
 <!-- #include "./params/ProcessName.md" -->
 
 ### -Top
 
-Specifies the maximum number to return.
+Specifies the maximum number to return.
```
DarqueWarrior, line 53: Update the help to say this is deprecated and going to be removed.

.docs/Get-VSTeamProcess.md
```diff
@@ -50,7 +62,10 @@ Default value: 100
 
 ### -Skip
 
-Defines the number of Process Templates to skip. The default value is 0
+Defines the number of Process Templates to skip.
```
DarqueWarrior, line 65: Update the help to say this is deprecated and going to be removed.

Source/Public/Get-VSTeamProcess.ps1
```diff
@@ -29,28 +30,25 @@ function Get-VSTeamProcess {
...
 
-            $objs = @()
-            
-            foreach ($item in $resp.value) {
-               $objs += [VSTeamProcess]::new($item)
-            }
-
-            Write-Output $objs
+           #we just fetched the processes so let's update the cache. Also Cache the URLS for processes
+            [VSTeamProcessCache]::processes = $resp.value.name | Sort-Object
```
DarqueWarrior: Something about updating the cache using the $Name filter seems like a bug. If the Name filter, filters the results to a single record that will be the only record in the cache. That would be a problem. All calls until the cache expires would only return that one record in the cache. The cache should not apply a Name filter.

DarqueWarrior: Would be nice if [VSTeamProcessCache]::Update was written to take an optional $list parameter. That way the $resp.value could be passed into [VSTeamProcessCache]::Update in this case. If Update is called with no parameters Get-VSTeamProcess would get called and update the cache with the full list of processes.

Looks like Update is expecting VSTeamProcess objects instead of the raw object from the service... So I guess you cloud

```
foreach ($item in $resp.value) {
   $objs += [VSTeamProcess]::new($item)
}

[VSTeamProcessCache]::Update -List $objs

return $objs |  Where-Object { $_.name -like $Name } | Sort-Object -Property Name
```

jhoneill: > Would be nice if [VSTeamProcessCache]::Update was written to take an optional $list parameter. That way the $resp.value could be passed into [VSTeamProcessCache]::Update in this case. 

Yes, that's the direction I'm going in. If these classes switch to C# then the content and timestamp would become hidden and you call "Invalidate" ,"IsExpired", "Update()" and/or "Update(content)".  I'm also moving towards the cache using Get-Process (not _getProcesses) but since that does a validate, which checks the cache it must ensure the validate returns when the parameter is empty without looking at the cache first otherwise it gets into an infinite loop (as I discovered!)

.docs/Get-VSTeamProcess.md
jhoneill: Will do.

Build-Module.ps1
```diff
@@ -83,6 +85,12 @@ if ($buildHelp.IsPresent) {
 Write-Output 'Publishing about help files'
 Copy-Item -Path ./Source/en-US -Destination "$output/" -Recurse -Force
 Copy-Item -Path ./Source/VSTeam.psm1 -Destination "$output/VSTeam.psm1" -Force
+if ($WithPublicClasses) {
```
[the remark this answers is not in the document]
jhoneill: Will do.  No one should need to hunt for the information.

jhoneill:
> Would be nice to apply the Update -List pattern to all cache classes.

See mu other comment - that's definitely the direction of travel.

Source/Private/common.ps1
```diff
@@ -154,63 +143,18 @@ function _getApiVersion {
    [CmdletBinding(DefaultParameterSetName = 'Service')]
    param (
       [parameter(ParameterSetName = 'Service', Mandatory = $true, Position = 0)]
-      [ValidateSet('Build', 'Release', 'Core', 'Git', 'DistributedTask', 'VariableGroups', 'Tfvc', 'Packaging', 'MemberEntitlementManagement', 'ExtensionsManagement', 'ServiceFabricEndpoint', 'Graph', 'TaskGroups', 'Policy')]
+      [ValidateScript({$_ -in ([VSTeamVersions] | Get-Member -Static  -MemberType Property).name})]
```
DarqueWarrior, line 146: Versions has more values like Account, DefaultProject and Version that are not valid for Service

DarqueWarrior:
How are you merging the changes from upstream?  I see 235 files changed but they really didn't it is making it very time consuming to review this PR.

Source/Private/common.ps1
jhoneill: Although they are not services, they can still be returned. If it's necessary to throw if someone uses _getApiVersion to find the account that can be done. 
However I am adding support for managing processes definitions which really needed its own API version  (possibly not with the most recent changes) and having to add to this list in common.ps1 when a change is made to class definition _felt_ wrong.

jhoneill:
> How are you merging the changes from upstream? I see 235 files changed but they really didn't it is making it very time consuming to review this PR.

It was some time back... I _think_ I pulled upstream into to my master branch and then merged master into this branch which had already been created for the PR.

DarqueWarrior:
Switching Get-VSTeamProcess from Processes to Work breaks TFS2017. I am working switching the call based on the API version being used. I will be merging these changes from another PR because there are too many files changed in this PR to make sense of the actual changes.

PR #344, opened 2020-07-21
Work item state commands

jhoneill:
# PR Summary

New commands to manage the states available to a WorkItem type in a custom process. (In the default processes the WorkItem types are fixed).  States are things like "New" "Done" etc. and the user can hide the built-in states, unhide them, and add or remove custom states.  So there 5 new commands 
`Get-VSTeamWorkItemState` gets a list of the states for a given WorkItem type
`Add-VSTeamWorkItemState` adds a custom type to the list
`Remove-VSTeamWorkItemState` removes a custom type the was previously added 
`Hide-VSTeamWorkItemState`  hides a system type from the list
`Show-VSTeamWorkItemState`  unhides a system type which was previously hidden

Each command has its own help file and unit test, and there is a format XML file to display the states as a table. 

## PR Checklist


Source/Private/common.ps1
```diff
@@ -152,58 +152,10 @@ function _getApiVersion {
    )
 
    if ($Target.IsPresent) {
-      return [VSTeamVersions]::Version
+         return [VSTeamVersions]::Version
    }
-   else {
+   else {return [VSTeamVersions]::$Service }
```
DarqueWarrior, line 157: This is an awesome change. I had no idea this would work.

DarqueWarrior:
Does this PR also have changes from 338? Some of these look familiar.

Source/Public/Add-VSTeamWorkItemState.ps1
```diff
@@ -0,0 +1,57 @@
+function Add-VsteamWorkItemState {
```
DarqueWarrior, line 1: A lot about this function makes me uncomfortable. Using -Url is really only to be used by Invoke-VSTeamRequest.

All functions should be able to call _callAPI using the proper Area and Resource parameters. 

I am also confused why you are using Add-Member instead of defining a class that is the shape of the object you want to return.

jhoneill: The first call which changes a system work item type to a custom one could use area and resource. 

The one which actually adds the state needs the whole URL of the work item type with "states" appended to it, 
To do it by passing the parts of the URI to _CallAPI, for assembly means "states" needs to be forced into the ID field to make it work. It made more sense to me at the time to say "We have the URI already, extend it and send it". 

Add-member is just the PowerShell way.  Something comes back as a custom object, and we a can extend that at will, or we can define a class and a constructor which copies everything over from the custom object to a new object and adds a couple of fields, but nothing which uses the states objects would see notice any difference between it being a custom object, and being a instance of a custom class.

DarqueWarrior: I am a big fan of consistency. When I read any function nothing should look odd or different if at all possible. It makes it much easier for others to collaborate when all the code "feels" the same. When I read code I should not be able to tell who on the team wrote it. It needs to look and feel like I wrote it. It makes me far more comfortable changing it.

jhoneill: I (mostly) agree. There is more than valid one way of doing a lot of this stuff, and if two ways do the job roughly as well as each other the "right" one is whichever makes for easier maintenance. But being open to contributions _and_ keeping stuff consistent (even if that means consistently _not_ following some normal practice or other) requires someone to put in the work to change things to match bits of style you didn't even realise were part of your style until someone did it another way. 

In some languages you can't have an object with a given set properties unless you define a class with those properties beforehand. But PowerShell creates custom objects as needed - we often return those from `Invoke-WebRequest`. Often the REST API returns an object without the information needed to fit it into the hierarchy (a state's work-item-type and the wit's Process template for example) which is needed when processing the object later. So the object from `Invoke-WebRequest` might be return by the module;  in script which uses the module might modify the object to identify its parent and grandparent: that's the expected use `Add-member`.
That's a common  situation - especially with the fields and work item layout stuff I have yet to send you - so I add the hierarchy information as part the function. Bringing paradigms of C# into PowerShell (explicit classes in this case) increases the amount of code - and spreads it over multiple files - but returns an object which the user / calling script thinks is identical. 

Requesting an object which fits in the hierarchy of  `Area/Resource/ID?version&Options` _and_ where know the parts is better done by passing the parts _callAPI.  If, on the other hand the URI follows some other pattern like WorkItemType/ID/Form/Pages/PageID/Groups/GroupID  and discovering the group (in this case) returns the URI, there is no need to invoke _callapi with 
```
$IDPart = regex to extract "ID/Form/Pages/PageID/Groups/GroupID" from $uri 
_callApi   -ID $IDPart
```
instead of 
`_callApi   -URL $uri`

However, if you feel better able to support the first one, that's still the way to go :-) 

J

PR #361, opened 2020-10-06
Extend  set of -WorkItemType commands

jhoneill:
# PR Summary

- Enhanced `Get-VSteamWorkItemType` , 
- New `Add-`, `Set-` & `Remove-`  `-VSTeamWorkItemType`  commands, including new C# Classes to support for Icon and Color parameters with completers and transformers.
- Enhanced `Set-VSTeamCurrentProject` to be Process-Template aware and added current process (process-template) to `Versions` class.  
** See extensive notes in section at the top of changelog.md**

## PR Checklist


jhoneill:
Scrub that comment, Azure Devops pipelines were misbehaving yesterday PM, and are back to normal now.

.docs/Add-VSTeamWorkItemType.md
```diff
@@ -0,0 +1,164 @@
...
+Aliases: Name
+
+Required: True
+Position: 1
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### CommonParameters
```
SebastianSchuetze: this is not needed as it will be generated automatically with the docs. You only need to document what you wrote and not common advanced cmdlet parameters or logic.

.docs/Get-VSTeamWorkItemIconList.md
```diff
@@ -0,0 +1,28 @@
...
+
+## SYNTAX
+
+## DESCRIPTION
+
+<!-- #include "./synopsis/Get-VSTeamWorkItemIconList.md" -->
+
+## PARAMETERS
+
+### CommonParameters
```
SebastianSchuetze: not needed, reason see above

.docs/Get-VSTeamWorkItemType.md
```diff
@@ -20,21 +20,78 @@
...
-This command gets a single work item type.
+This command gets a single WorkItem type from the named project.
+
+### Example 2
+
+```powershell
+Get-VSTeamWorkItemType -ProcessTemplate Basic
+```
+
+This command gets a list of the WorkItems available to projects which use the 'Basic' Template.
```
SebastianSchuetze: you mean work item types not WorkItems, right?

.docs/Remove-VSTeamWorkItemType.md
```diff
@@ -0,0 +1,130 @@
...
+Aliases: Name
+
+Required: True
+Position: 1
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### CommonParameters
```
SebastianSchuetze: not needed, reason see above

.docs/Set-VSTeamWorkItemType.md
```diff
@@ -0,0 +1,209 @@
...
+Aliases: Name
+
+Required: True
+Position: 1
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### CommonParameters
```
SebastianSchuetze: not needed, reason see above

Source/Public/Set-VSTeamDefaultProject.ps1
```diff
@@ -49,19 +49,19 @@ function Set-VSTeamDefaultProject {
 
    process {
       if ($Force -or $pscmdlet.ShouldProcess($Project, "Set-VSTeamDefaultProject")) {
-         if (_isOnWindows) {
```
SebastianSchuetze, line 24: can you explain in short the change in this cmdlet in regards to the process?

Source/Public/Set-VSTeamWorkItemType.ps1
```diff
@@ -0,0 +1,90 @@
...
+      [switch]$Force
+   )
+   process {
+      #Get the existing workitem-type definition. Drop out if can't be found 
+      $wit  = Get-VSTeamWorkItemType -ProcessTemplate $ProcessTemplate -WorkItemType $WorkItemType
+      if (-not $wit) {Write-Warning "'$workitemType' does not appear to be a valid Workitem type." ; return}
+      else           {$url  = $wit.url + "?api-version=" + (_getApiVersion Processes)  }
+
+#region set up the contents to go in the JSON body
+      $body = @{} #note. "Changing" some options to their current value has caused errors but those below seem OK...
```
SebastianSchuetze: is it needed that you have to post all the existing values? In some cases, the API is fine just posting what has changed.

Tests/function/tests/Add-VSTeamProject.Tests.ps1
```diff
@@ -88,6 +88,7 @@ Describe 'VSTeamProject' {
                Typeid = '00000000-0000-0000-0000-000000000002'
             }
          }
+         [vsteam_lib.ProcessTemplateCache]::Invalidate()
```
SebastianSchuetze, line 90: why do you need to invalidate the cache here? It is just a unit test, I thought that should not be needed.

Source/Public/Add-VSTeamWorkItemType.ps1
```diff
@@ -0,0 +1,46 @@
...
+
+   $body  = @{
+      name  = $WorkItemType
+      color = $Color
+      icon  = $Icon
+   }
+   if ($Description)  {$body['description'] =$Description }
+
+   if ($PSCmdlet.ShouldProcess($ProcessTemplate, "Add workitem '$WorkItemType' to process template"))  {
+      $resp = _callapi -Url $url -method  Post  -ContentType "application/json" -body (ConvertTo-Json $body) -erroraction stop
```
SebastianSchuetze: have you tried just handing over the body as an object? In the unit tests, you could then just assert by checking the properties of the psobject.

SebastianSchuetze: Also, you don't need to add content-type. By default, _callApi is using the UTF-8 content type for JSON, which should work better.

Source/Public/Add-VSTeamWorkItemType.ps1
```diff
@@ -0,0 +1,46 @@
...
+      if ($resp) {
```
SebastianSchuetze: when should there be not a response? If there is none shouldn't there be an exception? In which cases do you expect an empty response and not throw an exception?

SebastianSchuetze (changes requested):
I didn't have time to review all files. But I have some comments that need attention. :)

Source/Public/Set-VSTeamDefaultProject.ps1
jhoneill: The existing version sets an Environment variable "Team_Project" on windows but not on mac or linux.  The only part of this which should be in the _isOnWindows is persisting the value in the user or system environment, so the logic is now 
1. Set Team_project and Team_Process at process level on every OS.
2. If the OS is windows and $level is User or Machine , persist the variable at that level.

.docs/Set-VSTeamWorkItemType.md
jhoneill: Fixed

Source/Public/Set-VSTeamWorkItemType.ps1
jhoneill: In this case it is because a system-added WorkItemType being changed for the first time needs POST a request to create a new item inherited from it which can be customized. When I tested this just setting color, the server gave an error that it needed to have the icon, so I ensure all the fields are passed to the post, and rather than build the body two different ways - one a PATCH to a custom or inherited item and for a POST , I build the same body for both.

Tests/function/tests/Add-VSTeamProject.Tests.ps1
jhoneill: The cache is an object which is not cleared between different unit tests. So if test 1 assumes it is working with a custom process named "Scrum MkII" and only populates the cache with that ,and test 7 assumes it is working with the built in template "Agile" it will may "Agile is not a valid template" if the cache is not cleared. Because of the timing issue with the cache expiring it doesn't happen every time and took a little bit of debugging !

Source/Public/Add-VSTeamWorkItemType.ps1
jhoneill: I hadn't , I just automatically convert to JSON, and I hadn't seen inside the content type was set inside _callApi. Although I'm fairly sure I've copied that form from other functions, what you're describing would be better.

.docs/Get-VSTeamWorkItemType.md
jhoneill: yes. I'll change it.

Source/Public/Add-VSTeamWorkItemType.ps1
jhoneill: One place is in testing, if we only want to see the call was made and don't check what is done with result. Again it's mostly a habit / pattern. If something throws an error some things (the AD module for example) write a message to error but don't throw a terminating exception. There's a comment repeated in multiple places about  
>   I catch because using -ErrorAction Stop on the Invoke-RestMethod ... 
There the post api-call logic goes in the try block, and the error is echoed out in the catch block. I just let any message bubble up where it occurs, and only run the api-call logic if there was something to run it _on_. That's simpler / easier to read, I think.  It also works where you don't know if what your calling generates terminating errors or not. 

Here's a little bit of code 
```
function Test-one {
    [CmdletBinding()]
    Param ()
    Invoke-RestMethod "httttps://google.com"
    write-host "should not be executed in one"
}

function test-two {
    [CmdletBinding()]
    param()
    $x = test-one -ErrorAction stop
    "Two should not outout $x if 1 fails"
}
``` 
So here we call something with invoke-restmethod and a broken URI. There's code which shouldn't run if we get an error.  And `-ErrorAction stop` gives us what we want, the error, with its source,.  

```
function test-two {
    [CmdletBinding()]
    param()
    $x = test-one
    if ($x) {"Two should not outout $x if 1 fails"}
}
```
This version continues to run and does the write-host operation but doesn't run the last part of test-two because there is nothing in $x 
Finally 
```
function test-two {
    [CmdletBinding()]
    param()
    try {
      $x = test-one
      "Two should not outout $x if 1 fails" 
    } 
    catch {
        throw $_.exception
    }
}
``` 
Is longest, prevents the write-host from running, but tells us the error is at throw not where it really occurred.

Source/Public/Add-VSTeamWorkItemType.ps1
jhoneill: If I pass the body as a pscustomObject  -verbose shows the JSON looks like this  and I get an error
```
VERBOSE: Body @{description=Describes a divergence between required and actual behavior, and tracks the work done to correct the defect and verify the correction.; isDisabled=False; color=ff0000; icon=icon_insect}
````
if I do convertto-json it looks like this 
```
VERBOSE: Body {
  "description": "Describes a divergence between required and actual behavior, and tracks the work done to correct the defect and verify the correction.",
  "isDisabled": false,
  "color": "ff0000",
  "icon": "icon_insect"
}
```
Which has " marks and : instead of = 


But I have taken the `content-type.` out - I also made a change so set works with multiple names,

**Comment by DarqueWarrior** [comment]
All the code around $env:TEAM_PROCESS makes me very uncomfortable. If you want to set a default for it use 

```
$Global:PSDefaultParameterValues["*-vsteam*:ProcessTemplate"] = "yourTemplate"
```

ProjectName is an exception because it is used by 90% of the functions. ProcessTemplate is not.

File: Source/Public/Set-VSTeamDefaultProject.ps1 [hunk 11]
```diff
@@ -49,19 +49,19 @@ function Set-VSTeamDefaultProject {
...
+         #Set at the process level, for any OS
+         $env:TEAM_PROJECT = $Project
+         [vsteam_lib.Versions]::DefaultProject = $Project
 
-            # You always have to set at the process level or they will Not
-            # be seen in your current session.
-            $env:TEAM_PROJECT = $Project
-            [vsteam_lib.Versions]::DefaultProject = $Project
+         $env:TEAM_PROCESS = [vsteam_lib.Versions]::DefaultProcess = Get-VSTeamProcess  -ExpandProjects |
+               Where-Object Projects -Contains $Project | Select-Object -ExpandProperty Name
```
DarqueWarrior, line 29: As I mentioned in the comment of the PR I am not a fan of $env:TEAM_PROCESS. Even if I was this is not where it should be getting set. Why would anyone call Set-VSTeamDefaultProject and expect the Process to be set?

Source/Public/Set-VSTeamWorkItemType.ps1
```diff
@@ -0,0 +1,92 @@
+function Set-VSTeamWorkItemType {
+   [CmdletBinding(DefaultParameterSetName='LeaveAlone',SupportsShouldProcess=$true,ConfirmImpact='High')]
+   param(
+      [parameter(ValueFromPipelineByPropertyName=$true)]
+      [vsteam_lib.ProcessTemplateValidateAttribute()]
+      [ArgumentCompleter([vsteam_lib.ProcessTemplateCompleter])]
+      $ProcessTemplate = $env:TEAM_PROCESS,
```
DarqueWarrior, line 7: What makes this better than using the built-in PSDefaultParameterValues that the rest of the code uses for ProjectName?

Source/Public/Set-VSTeamDefaultProject.ps1
jhoneill: OK. We're going to have work though a difference in style and ideas of what's best practice. 
`PSDefaultParameterValues["*-vsteam*:Something"]`   Can be problematic 

If the code has 
`Param  ( [type]$something = [class]::member) `
or 
`Param  ( [type]$something = $Scope:Variable ) `
where scope is` Env: `, `Script:,` or` Global:` 
Then that variable will be set.  And we and can distinguished between _user-passed_ and _set-to-default_ by looking at `$PSBoundParameter.containskey('Something')`   _defaulted_, parameters are not _bound_ (btw the common trick of treating `$something` as a boolean to see if something was passed fails if the value converts to `false` (for example it's an integer or an Enum, and 0 is a valid value) so I commonly check for "was passed" this way.

Now: I have two ways I want to ask for WorkItem Types.  "Give me the types for the a project", (the current project by default) or "give me the types for a specific process template". (Scrum, agile etc)

This commandline should be impossible. 
`Get-VsteamWorkItemType  -ProjectName  MyScrumProject  -ProcessTemplate MyAgileTemplate`

So one would write 
```
Param (
   [Parameter(ParameterSetName='ByProject')]
   $ProjectName  
   [Parameter(ParameterSetName='ByTemplate')]
   $ProcessTemplate
```

So now the user can't type both. But now:
`Get-VsteamWorkItemType -ProcessTemplate MyAgileTemplate`
causes an error, because `PSDefaultParameterValues["*-vsteam*:ProjectName  "]`  **always** stuffs the current project name in. So in this case I have had to allow a parameter set which takes BOTH and ignores `ProjectName `if there is a user supplied  `ProcessTemplate` 

**Ideally Parameter defaults should be set where the parameter is defined** (and the help should say what the default is), there's obviously a big lump of work to change this for `ProjectName` parameters but I did want to get it right for new things.     
There are other places where it needs background knowledge to understand defaults are being set. @SebastianSchuetze  told me that _callapi has a default for contentType. It doesn't:
```
      [string]$OutFile,
      [string]$ContentType,
      [string]$ProjectName,
```
Buried in the function there is this 
```
      $params = $PSBoundParameters
      if ($false -eq $PSBoundParameters.ContainsKey("ContentType")) {
         $params.Add('ContentType', 'application/json; charset=utf-8')
      }
``` 
And on principle I shouldn't need to read the code inside a function to know to call it properly. 

To answer
> Why would anyone call Set-VSTeamDefaultProject and expect the Process to be set?

I don't think they would.  It is just **caching "Process of the current project".**  Otherwise when any command which works on process template is called either the user must supply the template name, or if no template is given the code needs to fetch all templates and find out which one the current project is in.  I **can** do that by `PSDefaultParameterValues["*-vsteam*:Something"]`  (which is still setting a default for process) if you understand the issue and still prefer that way I'm happy to comply and disagree.

Source/Public/Set-VSTeamWorkItemType.ps1
jhoneill: See long comment above :-)

Source/Public/Set-VSTeamDefaultProject.ps1
jhoneill: Additional it did seem unnecessary to add the default process template to the user or system environment variables,; It is set in the local environment when connecting to a project and a situation where the user changes one but not the other isn't good.. 
I'm also wondering what happens if the user changes `PSDefaultParameterValues["*-vsteam*:Something"] ` without calling `set-VSTeamDefaultProject. `

SebastianSchuetze:
@jhoneill one request: Could you tell us when your PRs are finished? You seem to be able to work on it in your work time. We are not. When we review something and you keep changing the files and adding things, you really make it hard for us to be willing to check the changes no matter how good they are.

Please finish the PR to a state where you not change anything anymore until we find the time to finish it.

jhoneill:
> @jhoneill one request: Could you tell us when your PRs are finished? You seem to be able to work on it in your work time. We are not. When we review something and you keep changing the files and adding things, you really make it hard for us to be willing to check the changes no matter how good they are.
> 
> Please finish the PR to a state where you not change anything anymore until we find the time to finish it.

Understood . This is the foundation for additional work for a stack of other stuff and goes back to what I put in in #320 which I posted 5 months ago today. I can freeze this branch and put things which are logically part of it in another branch. But then if you say "wouldn't it make sense to do X" I'd have to say well yes, that's in the PR which I put in after this one to fix stuff which you still haven't merged. And obviously that would be frozen, so it might be "it's in the follow-up to the follow-up PR". 

As well 3 new and two changed commands here (making Process information more accessible, and commands to add / change work item types)  I have 
2 for Add and change Process Templates and 5 for Process and Work Item behaviors (boards), and 2 for team settings which relate to behaviors - 9 in all in #364 which depend on this. 
5 which relate to the states you can set a work item to in  #362 which also depend on this. 
7 which relate to picklists, fields, and adding fields to a work item type in plus 6 new classes for picklists and fields in #370  which again depend on this. 
And about 10 more which handle layouts on Work Item types. (Adding/Changing/Removing Pages, Groups and controls) which depend on #370.
So a couple of dozen commands in five branches.  

Now... Doing the layout stuff I found every add and set might be working on a WorkItem type which was copied into the process template at creation, before changing it , an inherited type must be created because the original is locked. I'd copied the same code into multiple functions. So I pulled it out as Unlock-WorkItemType and pushed it down into this branch (and removed the duplicate code in four of the five branches)  I _could_ wait for you to finish this one, #362, #364, and #370; and then send a PR to that (and make a more complex set of dependencies). Or I can use the delay to put things right...

Source/Public/Add-VSTeamWorkItemType.ps1
jhoneill: I've taken that out. It needed a try-catch because if I try to add the same type to multiple Process templates, and it fails in the middle I don't want the pipeline to do some that were possible and leave other possible ones not done.

jhoneill:
@SebastianSchuetze 

> > @jhoneill one request: Could you tell us when your PRs are finished? You seem to be able to work on it in your work time. We are not. When we review something and you keep changing the files and adding things, you really make it hard for us to be willing to check the changes no matter how good they are.

**I'm calling this branch "Frozen" now.**.  I will only allow fixes for something to badly broken and to wait for a future PR. 
I've been using some quiet time at work.  We did a check and found we have over 1000 Azure devops organizations linked to our Azure AD: every contract creates an AzDO org and sets things up differently from every other project.  If you transfer between contracts you have to learn how the new one uses AzDO and management can't get usable data because everyone calls custom work item types different names, creates custom fields for the same purpose with different names and units. The end goal for all of this is "Azure DevOps as code" and projects which look very similar. 

Since all the other things I'm sending you are sub-branches of this one, it would be great (from my point of view) if you could progress this. I will confirm when the other ones are properly "frozen" as well. 
If I (and other people) seem to be taking the use of your personal time for granted then we owe you an apology. Hopefully there's some comfort in people seeing your work and assuming you must be working on it full time :-)

Source/Public/Set-VSTeamDefaultProject.ps1
DarqueWarrior: Sorry for the delay.  There is a lot to unpack here. I will try to take them one at a time. 

1. No matter if we decide to use environment variables or PSDefaultParametersValues there needs to be a Set-VSTeamDefaultProcessTemplate and Clear-VSTeamDefaultProcessTemplate. I don't want Set-VSTeamDefaultProject to have side effects.

2. I agree with you on the contentType default value on _callAPI and will address. 

3. Get-VsteamWorkItemType -ProjectName MyScrumProject -ProcessTemplate MyAgileTemplate should be impossible. I don't see how we set the default project affecting this. If we use the environment variable way or the PSDefaultParametersValues if a default is set for ProjectName you should get the error.

DarqueWarrior:
> @SebastianSchuetze
> 
> > > @jhoneill one request: Could you tell us when your PRs are finished? You seem to be able to work on it in your work time. We are not. When we review something and you keep changing the files and adding things, you really make it hard for us to be willing to check the changes no matter how good they are.
> 
> **I'm calling this branch "Frozen" now.**. I will only allow fixes for something to badly broken and to wait for a future PR.
> I've been using some quiet time at work. We did a check and found we have over 1000 Azure devops organizations linked to our Azure AD: every contract creates an AzDO org and sets things up differently from every other project. If you transfer between contracts you have to learn how the new one uses AzDO and management can't get usable data because everyone calls custom work item types different names, creates custom fields for the same purpose with different names and units. The end goal for all of this is "Azure DevOps as code" and projects which look very similar.
> 
> Since all the other things I'm sending you are sub-branches of this one, it would be great (from my point of view) if you could progress this. I will confirm when the other ones are properly "frozen" as well.
> If I (and other people) seem to be taking the use of your personal time for granted then we owe you an apology. Hopefully there's some comfort in people seeing your work and assuming you must be working on it full time :-)

I appreciate all the support but this is not my full-time job. Man, I wish it was. I am going to try and devote more time to it going forward but I appreciate your patience.

Source/Public/Set-VSTeamDefaultProject.ps1
DarqueWarrior: I am ok with finding an alternative to using PSDefaultParametersValues. So your method is fine but I still want separate functions for setting and clearing that value to be consistent with DefaultProject.

jhoneill: Don't worry about the delay. I worked at Microsoft a decade ago and know what it's like doing a big PowerShell thing as a side project (first PowerShell support for Hyper-V, when 2008 didn't have a module for it... that was me).   

1. I understand the not wanting side effects thing, but the idea is to have the default process template as _process for the current project_ and I'm just caching it instead of querying for it.  
Just checking my thinking the default process template _should_ follow the default project , shouldn't it ?  It would be quite easy don't change the process when the project changes under some conditions. 
 
In a way it would be better to have a `-DefaultProject` parameter for `Set-VSTeamAccount` so the job is done in one command instead of two/three and then have setting the project or process as individual commands which only changed one variable.   

2. :-)  

3. That's what most people would expect. But PSDefaultParameterValues  _fakes_ a command line parameter.  Here's a little sample
```
function test {
[cmdletbinding(DefaultParameterSetName="one")]
param (
    [parameter(mandatory=$true, ParameterSetName="one")]
    $param_one ,
    [parameter(ParameterSetName="two")]
    $param_two
)
$PSBoundParameters
$param_one
$param_two
}
```
on its own  `test`  prompts
`Supply values for the following parameters:`   
`param_one: `   

`test -param_two 2` or   `test -param_one 1` work as expected and `test -param_one 1 -param_two 2` throws a parameter error. If you then do 
`$PSDefaultParameterValues['Test:param_one']="Alpha" `     
`test`  works with no parameters on the command line and reports that `Param_one` was a bound parameter. 

What Default Parameter Values does is to fake  `-param_one "Alpha"` being on the commandline so if you type    
`test -param_two 2`  the command will always see `test -param_two 2 -param_one "Alpha"` which isn't valid. 

But if you write the function like this :
```
function test2 {
[cmdletbinding(DefaultParameterSetName="one")]
param (
    [parameter(ParameterSetName="one")]
    $param_one = $global:Par1, 
    [parameter(ParameterSetName="two")]
    $param_two
)
$PSBoundParameters
$param_one
$param_two
}
```
`Test2` will prompt for input unless you do  `$global:par1 = "Alpha"` first  and it works with either parameter but not both. 
However when the value is defaulted in, it is _not a bound parameter_ so `test` on its own shows $param_one is "Alpha" but there are no bound parameters, and `test -param_two  2` shows $param_two has a value, `psBoundParameters` contains Param_two = 2 and the **variable**  $param_one has its default value even though there is no **parameter** param_one (it's the opposite of dynamic parameters, they go in PSBoundParameters but aren't automatically variables). 

I can think of 3 reasons for avoiding using `PSDefaultParameterValues` with wildcards - putting the parameter in when it isn't wanted is one, not having something in the function's parameter block telling you how a default gets there is a second (see #2 :-)  ) and another script might reset it.

DarqueWarrior: > In a way it would be better to have a `-DefaultProject` parameter for `Set-VSTeamAccount` so the job is done in one command instead of two/three and then have setting the project or process as individual commands which only changed one variable.

Adding a -DefaultProject parameter to Set-VSTeamAccount you would lose tab completion and validation because you have not authorized the account.

DarqueWarrior: Now we could add Set-VSTeamDefault and that way we only change that one function as we discover new defaults we want to provide. We need to leave Set-VSTeamDefaultProject for a while to allow a transfer from one to the other.

jhoneill: > Adding a -DefaultProject parameter to Set-VSTeamAccount you would lose tab completion and validation because you have not authorized the account.

Yes, but I think that would be OK. Currently I have 
```
Set-VSTeamAccount -Account $account  -SecurePersonalAccessToken (Import-Clixml .\myToken.xml) ;
Set-VSTeamDefaultProject $projectName
```

In a file 
And I'm trying to avoid 
`Set-VSTeamDefaultProcess ((Get-VSTeamProcess -ExpandProjects) | where projects -Contains $projectname)`

In an ideal world I'd have a utility to create _credential_ where the user name was Account\project and the password was the access token (utility would get this from the clipboard so it's never in the history) and export it with `export-cliXML`   to vsteamCred.xml in my home directory  (a secure string saved like that can only be decoded by the person who saved it)  then  
`Connect-VSTeam  -credFile mycred.xml) ` 
using import-clixml to get the cred,  splitting name at the \ to get the account details and logging on using the securestring password (accesstoken) as now.    
If there is anything after  \ in the account name it calls `Set-VSTeamDefaultProject` with that.     
And either `Set-VSTeamDefaultProject` or a call to new code in either in a new `Set-VSTeamDefaultProcess` or inside this `connect` command would  set the process default.  
As a default I'd have `home\vsteamCred.xml` as the path so can just run Connect-Vsteam (or some alias). And for those who don't want a credential file even with secure strings I'd allow credential input so they can paste a plain text access token but not into the command line. 

If the user is logged on when they  create the cred file, the account is known, they can use the current default project or specify one.

DarqueWarrior:
I will be returning to this PR soon. I used the $env:XXX for defaults in my last PR so I will work on changing the way the current defaults are implemented to match.

DarqueWarrior:
Hey, @jhoneill would you be open to a Teams meeting to work through this PR? There are some design decisions I want to make sure I understand. I want the contributions but I also want them to be consistent with the rest of the module. I am open to being convinced to change the rest of the module if that is the correct move.

SebastianSchuetze:

PR #362, opened 2020-10-06
Commands to work with states available for WorkItem types

jhoneill:
# PR Summary
Follow on from #361 
 Add support for workitem states    
*  Get-VsteamWorkItemState    
*  Add-VSTeamWorkItemState   
*  Remove-VsteamWorkItemState  
*  Hide-VsteamWorkItemState    
*  Show-VsteamWorkItemState 
and associated Format file + tests and help

Each WorkItem-type in each process-template has a set of possible states e.g. "To Do", "In Progress" and "Done". Some are system states which cannot be removed but can be hidden (and shown). Other states are custom/user-added and can be removed (but not hidden). This PR provides tools for listing the states, hiding and showing system ones, and adding and removing user ones. 

## PR Checklist

- [X] [Write Help](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#write-help)
- [X] [Write Unit Test](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#write-unit-test)
- [X] [Update CHANGELOG.md](https://github.com/DarqueWarrior/vsteam/blob/master/.github/CONTRIBUTING.md#update-changelogmd)

PR #364, opened 2020-10-12
Extra process commands

jhoneill:
# PR Summary

1.  Added Commands `Add-` and `Set-` `-VSTeamProcess`. `Add-` creates a process template based on a pre-existing one, Set- enables/disables, sets the default for new projects and changes the name / description.
2.  Added Commands `Get-` `Set-` `Add-` `-Remove` `-VSTeamProcessBehavior` and a format XML file for process behaviors. (a.k.a boards). Combined with (1) allows creation of a custom process template with its own boards. 
3.  Added Commands `Get-` `Set-` `-VSTeamWorkItemBehavior` and a format XML file  for WorkItem behaviors. These work with boards from (2) or with a custom-templates built-in boards to add work item types to a board. 
4.  Changed command `Get-VSTeamIteration` adding a switch -Expand to return all the iterations returned by the -Depth option, rather than one with a children node where each element may have children. 
5.  Added Commmands Get- and Set- -VSTeamSetting - to allow bug behavior and visibile boards (and things set with them) to be set. Otherwise the boards from (2) can't be seen. 

## PR Checklist

Test count now at 735. Tests for settings are fairly simple, others more sophisticated. Text above is in the change log. Note that this depends on  #361 but not on #362

### DIFF
--- a/.docs/Add-VSTeamProcess.md
+++ b/.docs/Add-VSTeamProcess.md
@@ -1,0 +1,83 @@
+<!-- #include "./common/header.md" -->
+
+# Add-VSTeamProcess
+
+## SYNOPSIS
+
+<!-- #include "./synopsis/Add-VSTeamProcess" -->
+
+## SYNTAX
+
+## DESCRIPTION
+
+Add-VSTeamProcess will add a new Process template to your organization. Processes define the available WorkItem Types and the behaviours (backlog levels) for projects. Existing projects can be switched to a new process, and any enabled process can be used for creating new projects.
+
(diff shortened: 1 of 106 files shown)


SebastianSchuetze:

jhoneill:
@SebastianSchuetze  yes, probably best from a house keeping point of view that this (and the others you've closed) are closed. It's a pity that I did so much work that couldn't be used, in part because I had done so much 
I have some down time at the moment and might see if any of it can be salvaged, but I haven't looked at this code in over two years.

PR #366, opened 2020-10-13
Fix #365

### LINKED ISSUE #365: Error accessing [vsteam_lib.QueryCache]
Steps to reproduce
------------------

```powershell
[vsteam_lib.QueryCache]::GetCurrent()
```

Expected behavior
-----------------
List of queries

Actual behavior
---------------

```none
MethodInvocationException: Exception calling "GetCurrent" with "0" argument(s): "A command is required to add a parameter. A command must be added to the PowerShell instance before adding a parameter."
```

I think this code 
```C#
   list = Cache.Shell.AddArgument(queries)
                                 .AddCommand("Select-Object")
                                 .AddParameter("ExpandProperty", "Name")
                                 .AddCommand("Sort-Object")
                                 .Invoke<string>();

```
will always cause that error.  This works. 
```c#
list = Cache.Shell.AddCommand("Select-Object")
                                 .AddParameter("ExpandProperty", "Name")
                                 .AddCommand("Sort-Object")
                                 .Invoke<string>(queries);
```

jhoneill:
# PR Summary

It appears this 
```
               list = Cache.Shell.AddArgument(queries)
                                 .AddCommand("Select-Object")
                                 .AddParameter("ExpandProperty", "Name")
                                 .AddCommand("Sort-Object")
                                 .Invoke<string>();
```
Is the wrong way to pass `queries` down the pipeline into `select-object` and causes 
`A command is required to add a parameter. A command must be added to the PowerShell instance before adding a parameter.`

It's fixed by changed to calling the command like this.
```
list = Cache.Shell.AddCommand("Select-Object")
                                 .AddParameter("ExpandProperty", "Name")
                                 .AddCommand("Sort-Object")
                                 .Invoke<string>(queries);
```

## PR Checklist

### DIFF
--- a/Source/Classes/Cache/QueryCache.cs
+++ b/Source/Classes/Cache/QueryCache.cs
@@ -36,11 +36,10 @@ namespace vsteam_lib
                }
 
                // This will return just the names
-               list = Cache.Shell.AddArgument(queries)
-                                 .AddCommand("Select-Object")
+               list = Cache.Shell.AddCommand("Select-Object")
                                  .AddParameter("ExpandProperty", "Name")
                                  .AddCommand("Sort-Object")
-                                 .Invoke<string>();
+                                 .Invoke<string>(queries);
             }
          }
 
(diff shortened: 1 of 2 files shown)


DarqueWarrior:
Do you have an example where you get that error? I have not seen that error.

jhoneill:
> Do you have an example where you get that error? I have not seen that error.

The completer for queries doesn't use the cache, so I don't think _anything_ is calling it, meaning the error isn't visible - I can't recall why I tried to the access the cache directly.

PR #370, opened 2020-10-16
Work item field commands

jhoneill:
# PR Summary

1. Added Classes to support tab completion of Fields and Picklists in the following functions. Classes are:
    - FieldCache 
    - FieldCompleter 
    - FieldTransformAttribute
    - PicklistCache
    - PickListCompleter
    - PickListTransformAttribute 
2. Added functions Get-VSTeamWorkItemField and Add-VSTeamWorkItemField to list the fields in and add fields to a given WorkItem Type,
3. Added functions Get-VSTeamField and Add-VSTeamField to list global fields which can be used in (2) and define new ones.
4. Added functions Get-VSTeamPickList, Add-VSTeamPickList and Set-VSTeamPickList to list and modify picklists used in the fields in (3) and define new ones

This builds on #361 

## PR Checklist


bmkaiser:
Did these changes get merged in some other way? The documentation for `Add-VSTeamWorkItem` states that the `-WorkItemType` parameter should be able to use tab completion, but I find that nothing happens when I try even though I've set a default project. I believe the changes that are part of this PR would address this.

SebastianSchuetze:

PR #375, opened 2020-11-30
(maint) Fix typo in example

glennsarti:
# PR Summary

Previously the example used the plural of the cmdlet name, however the actual
name is just singular.  This commit updates the example text to use the correct
name and removes trailing whitepsace.

## PR Checklist

N/A for a docs only change.


SebastianSchuetze:
thanks for spotting and correcting it!

Issue #373, opened 2020-11-09
Get-VSTeamProject (and the cache) still only returns the top 100 by default


ghvanderweg:
It seems to me that #363 wasn't fixed in #367. 

Get-VSTeamProject still only returns the first 100 projects, and only those projects are saved in the cache.

Adding `Top` to the `PSDefaultParameterValues` helps with calling Get-VSTeamProject, but it doesn't increase the cache size. Trying to get a project by name that's not in the first 100 projects still fails validation.

Steps to reproduce
------------------
Assuming an account with 129 projects.

```powershell
(Get-VSTeamProject).Count
$PSDefaultParameterValues.Add("*-VSTeam*:Top",1000)
(Get-VSTeamProject).Count
Get-VSTeamProject -Name 'TestProject101'

```

Expected behavior
-----------------

```none
129
129

Name         Description
----         -----------
TestProject101
```

Actual behavior
---------------

```none
100
129
Get-VSTeamProject: Cannot validate argument on parameter 'Name'. 'TestProject101' is invalid
```

Environment data
----------------

OS

- [ ] macOS
- [x] Windows
- [ ] Linux

Server

- [ ] TFS 2017
- [ ] TFS 2018
- [ ] Azure DevOps Server
- [x] Azure DevOps Service

```powershell
> Get-VSTeamAPIVersion
Name                           Value
----                           -----
HierarchyQuery                 5.1-preview
Build                          5.1
DistributedTaskReleased        5.1
ServiceEndpoints               5.0-preview
Graph                          6.0-preview
Release                        5.1
Tfvc                           5.1
TaskGroups                     6.0-preview
MemberEntitlementManagement    6.0-preview
Pipelines                      5.1-preview
Git                            5.1
ExtensionsManagement           6.0-preview
VariableGroups                 5.1-preview.1
Packaging                      6.0-preview
Core                           5.1
Version                        VSTS
DistributedTask                6.0-preview
Policy                         5.1
Processes                      6.0-preview
```

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      7.0.3
PSEdition                      Core
GitCommitId                    7.0.3
OS                             Microsoft Windows 10.0.19041
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

[...]

PR #377, opened 2020-12-02
Addressing Issue #373

DarqueWarrior:
# PR Summary

Azure DevOps only returns 100 projects unless you override the top value. This PR allows you to set a default value of projects to return. This will allow tab completion and validation to work on all projects. 

## PR Checklist

### DIFF
--- a/Source/Public/Set-VSTeamDefaultProjectCount.ps1
+++ b/Source/Public/Set-VSTeamDefaultProjectCount.ps1
@@ -8,35 +8,7 @@ function Set-VSTeamDefaultProjectCount {
       [int] $Count
    )
    DynamicParam {
-      $dp = New-Object System.Management.Automation.RuntimeDefinedParameterDictionary
-
-      # Only add these options on Windows Machines
-      if (_isOnWindows) {
-         $ParameterName = 'Level'
-         # Create the collection of attributes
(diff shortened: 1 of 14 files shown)


Source/Public/Set-VSTeamDefaultProjectCount.ps1
```diff
@@ -0,0 +1,63 @@
...
+      [switch] $Force,
+
+      [Parameter(Mandatory = $true, ValueFromPipelineByPropertyName = $true)]
+      [int] $Count
+   )
+   DynamicParam {
+      $dp = New-Object System.Management.Automation.RuntimeDefinedParameterDictionary
+
+      # Only add these options on Windows Machines
+      if (_isOnWindows) {
```
SebastianSchuetze: aren't you creating duplicate code here. the clear function has the same. Is it possible to group the dp creation?

SebastianSchuetze (commented):
looks good to me. Just one minor issue.

Source/Public/Set-VSTeamDefaultProjectCount.ps1
DarqueWarrior: Good point. It is actually in four places now. Let me see about refactoring this.

DarqueWarrior: I think the code is much better now.

Issue #218, opened 2019-11-08
Support for creating, configuring and removing Agent pools


rwaal:
I would like to be able to create and configure Agent Pools in Azure DevOps. This would include things such as:

- Creating a new Agent pool.
- Configure settings, such as auto-provioning and maintenance jobs.
- Configure the security settings on the Agent pools.
- Removing Agent pools.

I looked in the available cmdlets and it seems the module doesn't provide the functionality. I am new to the module, so perhaps I overlooked. Any insights on this would be appreciated. Thanks allot!

SebastianSchuetze:
Hey, nice idea. Let me check that I can provide you with some cmdlets. 

I made a quick research and found the following documented APIs to create cmdlets from:

* [Managing Agent Pools](https://docs.microsoft.com/en-us/rest/api/azure/devops/distributedtask/pools?view=azure-devops-rest-5.1) for adding, deleting, updating and retrieving
* [Connecting pools with projects](https://docs.microsoft.com/en-us/rest/api/azure/devops/distributedtask/queues?view=azure-devops-rest-5.1) which makes an agent pool available to a project by adding, deleting and retrieving

The followings do not have a documented API as far as I can see:
* configure maintenance jobs
* configure permissions

rwaal:
Thanks for the quick response! It would be awesome if these things could be integrated in de module. 

Like I mentioned, I’m new to the module. But please let me know if there’s anything I can do to help.

DarqueWarrior:
Given the links above you should be able to call the API with the module today. Just use the

```Powershell
Invoke-VSTeamRequest
```

Invoke-VSTeamRequest can all any API even if we have not had time to wrap it.

SebastianSchuetze:
Working on this now. I will start with adding, updating and removing agent pools.

SebastianSchuetze:
I started implementing the following things and want them to be merged first

* delete
* add
* update

Agent pools and their basic settings like auto-provision, description, name, auto-update

SebastianSchuetze:
@rwaal you might want to check the connected PR to see if that fulfils your request.

rikribbers:
Hi, nice to see this is going somewhere. Will this include support for VM scalesets as well?

SebastianSchuetze:
Probably not yet (I forgot what I implemented). The PR is waiting for Donovans Review. But it can be implemented afterwards.

SebastianSchuetze:
[...]

PR #384, opened 2021-03-07
Agent Pools cmdlets for removing, adding and updating

SebastianSchuetze:
# PR Summary

implemented idea #218 

## PR Checklist


SebastianSchuetze:
@DarqueWarrior Unit tests past. Hope everything is good. But the cmdlets are straight forward.
Don't close the related issue #218 , as I want this first in merged and then add some more a bit later.

Source/Classes/Completer/BaseCompleter.cs
```diff
@@ -11,40 +11,71 @@
...
+
+        private static CompletionResult GetCompletionResult(string word, string wordLabel, CompletionResultType resultType)
+        {
+            // Only wrap in single quotes if they have a space. This makes it easier
+            // to use on macOs
+            return (new CompletionResult(word.Contains(" ") ? $"'{word}'" : word, wordLabel, resultType, wordLabel + ": " + word));
+
+        }
+
+        protected static void SelectValues(string wordToComplete,
```
SebastianSchuetze, line 62: I basically added here a method to allow completer to have display names of completers to be different from values.

Source/Classes/Completer/BaseCompleter.cs
```diff
@@ -11,40 +11,71 @@
...
+            }
+        }
+
+
+        private static CompletionResult GetCompletionResult(string word)
+        {
+            return GetCompletionResult(word,word,CompletionResultType.Text);
+        }
+
+        private static CompletionResult GetCompletionResult(string word, string wordLabel, CompletionResultType resultType)
```
SebastianSchuetze, line 54: For the updated completer for time zones I added a completion result method with two different method headers

Source/Classes/TimeZones.cs
```diff
@@ -0,0 +1,165 @@
...
+
+namespace vsteam_lib
+{
+   public static class TimeZones
+   {
+
+      public static Dictionary<string, string> GetTimeZones()
+      {
+
+         var timeZones = new Dictionary<string, string>(){
```
SebastianSchuetze, line 13: I took these values straight from Azure DevOps UI. They are used everywhere where you have time zones (e.g. pool maintenance, user profile time& region etc.). So this is generalized.

image

Source/Public/Set-VSTeamAgentPoolMaintenance.ps1
```diff
@@ -0,0 +1,97 @@
...
+      if ($force -or $pscmdlet.ShouldProcess($Id, "Set Pool Maintenance")) {
+
+         $isEnabled = $true
+         if ($Disable.IsPresent) {
+            $isEnabled = $false
+         }else {
+
+         }
+
+         $resp = _callAPI -Method Get -NoProject -Area distributedtask -Resource pools -Id "$Id/maintenancedefinitions" -Version $(_getApiVersion DistributedTask)
```
SebastianSchuetze, line 49: @DarqueWarrior  hope you don't mind. That I made this cmdlet to handle adding and updating a maintenance schedule in one command. Technically you can add only one, but there are only two states:

* no existing schedule --> add one
* update existing schedule

So I made it as you can just set the schedule. Also, activated/deactivated the schedule to follow the same experience as in the GUI, which makes more sense to me.

SebastianSchuetze (commented):
@DarqueWarrior can you have a look here. It's a bunch of files, but I tried to follow your lead here. I still need to add unit tests for the pool maintenance cmdlets

PR #400, opened 2021-06-26
Fixes ambiguity of parameter sets in VSTeamUserEntitlement cmdlets

### LINKED ISSUE #393: Ambiguity or bug in Update-VSTeamUserEntitlement
Steps to reproduce
------------------

```powershell
update-VSTeamUserEntitlement -id $userDescriptor -license Stakeholder
```

Expected behavior
-----------------

The license for the specified user should be updated. It can be updated by using the following format
```powershell
Get-VSTeamUserEntitlement -Id $userDescriptor|Update-VSTeamUserEntitlement -License StakeHolder
```

Actual behavior
---------------

```none
Update-VSTeamUserEntitlement: Parameter set cannot be resolved using the specified named parameters. One or more parameters issued cannot be used together or an insufficient number of parameters were provided.
```
There's no detail in the get-help documentation for this particular cmdlet specifying any examples, other than to indicate that the combination of id and license should be valid.  

It's also unclear as to what input that this command is accepting via the pipeline in the second example which does work.  

Environment data
----------------

```powershell
> Get-VSTeamAPIVersion


Billing                     : 5.1-preview.1
Build                       : 5.1
Core                        : 5.1
DistributedTask             : 6.0-preview
DistributedTaskReleased     : 5.1
ExtensionsManagement        : 6.0-preview
Git                         : 5.1
Graph                       : 6.0-preview
HierarchyQuery              : 5.1-preview
MemberEntitlementManagement : 6.0-preview
Packaging                   : 6.0-preview
Pipelines                   : 5.1-preview
Policy                      : 5.1
Processes                   : 6.0-preview
Release                     : 5.1
ServiceEndpoints            : 5.0-preview
TaskGroups                  : 6.0-preview
Tfvc                        : 5.1
VariableGroups              : 5.1-preview.1
Version                     : VSTS

```

```powershell
> $PSVersionTable
PSVersion                      7.1.3
PSEdition                      Core
GitCommitId                    7.1.3
OS                             Linux 5.10.27-gentoo-x86_64 #1 SMP Fri Apr 2 16:48:21 BST 2021
Platform                       Unix
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}

```

SebastianSchuetze:
- fixes #393
- added tests for changed parameter

- [ ] Update Changelog and Moduleversion

### DIFF
--- a/Source/Public/Update-VSTeamUserEntitlement.ps1
+++ b/Source/Public/Update-VSTeamUserEntitlement.ps1
@@ -1,32 +1,28 @@
 function Update-VSTeamUserEntitlement {
-   [CmdletBinding(SupportsShouldProcess = $true, ConfirmImpact = "High", DefaultParameterSetName = 'ByEmailLicenseOnly',
-    HelpUri='https://methodsandpractices.github.io/vsteam-docs/docs/modules/vsteam/commands/Update-VSTeamUserEntitlement')]
+   [CmdletBinding(SupportsShouldProcess = $true, ConfirmImpact = "High", DefaultParameterSetName = 'ByEmail',
+      HelpUri = 'https://methodsandpractices.github.io/vsteam-docs/docs/modules/vsteam/commands/Update-VSTeamUserEntitlement')]
    param (
-      [Parameter(ParameterSetName = 'ByIdLicenseOnly', Mandatory = $True, ValueFromPipelineByPropertyName = $true)]
-      [Parameter(ParameterSetName = 'ByIdWithSource', Mandatory = $True, ValueFromPipelineByPropertyName = $true)]
+      [Parameter(ParameterSetName = 'ById', Mandatory = $True, ValueFromPipelineByPropertyName = $true)]
       [Alias('UserId')]
       [string]$Id,
 
-      [Parameter(ParameterSetName = 'ByEmailLicenseOnly', Mandatory = $True, ValueFromPipelineByPropertyName = $true)]
-      [Parameter(ParameterSetName = 'ByEmailWithSource', Mandatory = $True, ValueFromPipelineByPropertyName = $true)]
+      [Parameter(ParameterSetName = 'ByEmail', Mandatory = $True, ValueFromPipelineByPropertyName = $true)]
       [Alias('UserEmail')]
       [string]$Email,
 
-      [Parameter(ParameterSetName = 'ByIdLicenseOnly', Mandatory = $true)]
-      [Parameter(ParameterSetName = 'ByIdWithSource')]
-      [Parameter(ParameterSetName = 'ByEmailLicenseOnly', Mandatory = $true)]
-      [Parameter(ParameterSetName = 'ByEmailWithSource')]
+      [Parameter(ParameterSetName = 'ById', Mandatory = $false)]
+      [Parameter(ParameterSetName = 'ByEmail', Mandatory = $false)]
       [ValidateSet('Advanced', 'EarlyAdopter', 'Express', 'None', 'Professional', 'StakeHolder')]
       [string]$License,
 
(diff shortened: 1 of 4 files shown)


Tests/function/tests/Update-VSTeamUserEntitlement.Tests.ps1
```diff
@@ -16,15 +16,33 @@ Describe "VSTeamUserEntitlement" -Tag 'VSTeamUserEntitlement' {
...
 
-      It 'by email should update a user' {
+      It 'by email without license should update a user' {
+         Update-VSTeamUserEntitlement `
+            -LicensingSource msdn `
+            -MSDNLicenseType enterprise `
+            -Email 'dlbm3@test.com' `
+            -Force
+
+         Should -Invoke _callAPI -Exactly -Times 1 -Scope It -ParameterFilter {
```
usepowershell, line 34: There are three values that are being passed up to the API in each of these requests.  The only thing the function does is determine if there's a new value to apply or just pass back the current value.

This test should assert that the submitted update object has the expected license type since that isn't changing per the update call.

Tests/function/tests/Update-VSTeamUserEntitlement.Tests.ps1
```diff
@@ -35,12 +53,15 @@ Describe "VSTeamUserEntitlement" -Tag 'VSTeamUserEntitlement' {
...
+            $body -like '*"accountLicenseType":"Stakeholder"*' -and
+            $body -like '*"licensingSource":"msdn"*' -and
+            $body -like '*"msdnLicenseType":"enterprise"*' -and
             $resource -eq 'userentitlements' -and
             $version -eq $(_getApiVersion MemberEntitlementManagement)
          }
       }
 
-      It 'by id should update a user' {
+      It 'by id without license should update a user' {
```
usepowershell, line 64: This test is missing similar validations to `by email without a license`.

Source/Public/Update-VSTeamUserEntitlement.ps1
```diff
@@ -69,7 +65,7 @@ function Update-VSTeamUserEntitlement {
          }
       }
 
-      $body = ConvertTo-Json -InputObject @($obj) -Depth 100
+      $body = ConvertTo-Json -InputObject @($obj) -Depth 100 -Compress
```
usepowershell, line 68: I can't put this comment down below by the write-outputs, but I'd consider Write-Information or Write-Host or create a structured object to hold the new and old values.  Write-output puts that text on the pipeline and may lead to unexpected behavior in a pipeline.  Another option would be a `-passthru` parameter to pass the updated object along the pipeline.

Probably not a change for now, but worth considering for future releases.

usepowershell (approved):
This change shouldn't break existing behavior unless someone was counting on the buggy parameter validation experience to throw.

This change should bring the actual behavior in line with the documented behavior.

Are there any docs that need to change with the parameter set change?

SebastianSchuetze:
Thanks and good point. I gonna check the docs for this.

Source/Public/Update-VSTeamUserEntitlement.ps1
SebastianSchuetze: changed this now to Write-Information

Tests/function/tests/Update-VSTeamUserEntitlement.Tests.ps1
SebastianSchuetze: @smurawski I really tried to understand what you are saying. But I don't get what I should test in your opinion. The licenses are asserted, arent't they?

the cdmdlet does not have return so I can't test the correct return values. I will go ahead with the bug fix, but maybe you could still comment on this and I can improve the test later.

PR #427, opened 2021-11-18
Update integration tests pipeline to use pipeline artifact

brittandeyoung:
# PR Summary

Updated Integration tests to pull down the module code from the pipeline artifact and install it. This eliminates the need to have the external nuget feed dependency. As I am unable to trigger pipeline runs for this project, I am unable to test this. 

I was able to download the module artifact and test this install method locally on my machine without issue.

Please review and let me know what you think.

Closes# 424

### DIFF
--- a/.build/integrationTests.yml
+++ b/.build/integrationTests.yml
@@ -28,6 +28,10 @@ jobs:
               displayName: 'Download tests'
               artifact: test
 
+            - download: current
+              displayName: 'Download module'
+              artifact: module
+
             - task: AzureResourceGroupDeployment@2
               displayName: 'Start TFS 2017'
               inputs:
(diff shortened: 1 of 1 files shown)


SebastianSchuetze:
@brittandeyoung this is not working, since the module has two dependencies that need to be installed as well.
Could you try to use the following approach of a local repository config and install?

https://stackoverflow.com/questions/49987884/how-to-install-update-a-powershell-module-from-a-local-folder-set-up-an-intern

brittandeyoung:
@SebastianSchuetze I have moved my efforts over to the GitHub actions fix. I will be abandoning this PR. I have the dependencies working in that branch which has an open PR for you. https://github.com/MethodsAndPractices/vsteam/pull/429

Issue #438, opened 2022-01-12
❗❗📢  Removal of support of TFS 2017 and 2018 with VSTeam version 8.0.0


SebastianSchuetze:
The plan is the remove support TFS 2017 and 2018. We have no possibility to test this and is cluttering the code base. This is getting harder to manage.

With older version VSTeam can still be used or by setting the VSTeam API manually or using `Invoke-VSTeamRequest`.

It's not set yet when this will happen (few months probably). I will add a warning message with the next release that will be shown when the module is loaded.

SebastianSchuetze:
I am starting to work on this and will be merged an extra branch for Version 8.0.0 until everything is ready.

okamilobo:
@SebastianSchuetze we started seeing this warning, but are having trouble figuring out is causing us to get the warning.  Could you please elaborate on what we should be looking for in our code and what to upgrade to?

PR #440, opened 2022-01-12
Added breaking change warning and new version notification on module load

SebastianSchuetze:
Added notification message if a new version is abailable.

Added warning message for upcoming changes in #438


SebastianSchuetze:
@DarqueWarrior what do you think of this?

I found a way that is cheaper than checking PS Gallery. We go directly to GitHub and that is fast enough to check on every module load:

image

SebastianSchuetze:
Also do you think the warning message is enough or should I not add this hard coded?

SebastianSchuetze:
I just need to find a way that this is not called during unit tests. Seems to stress the rate limit of the GitHub API too much. 😆

DarqueWarrior:
Have you tried putting this in the psm1 file instead of calling it on every web request that should only be run when the module is loaded.

SebastianSchuetze:
It is in the psm1 file 🙂. But unit tests seem to load the whole module several times.

But I got an idea. Maybe I offer an option with an environment variable that is checked with boolean.

So people can turn of the warning. And also the update notification.

PR #459, opened 2022-03-02
enable API version 6.0 for Get-VSTeamUserEntitlement

### LINKED ISSUE #388: Get-VSTeamUserEntitlement only returns 96 results
Steps to reproduce
------------------

```powershell
$users = Get-VSTeamuser -SubjectTypes aad
$userentitlements = Get-VSTeamUserEntitlement

```

Expected behaviour
-----------------

```none
$users.count
359

$users | ? mailaddress -eq "<my email address>"
<returns information>


$userentitlements.count
359
$userentitlements | ? displayname -eq "<my first name>*"
<returns information>
```

Actual behaviour
---------------

```none
$userentitlements.count
96
$userentitlements | ? displayname -eq "<my first name>*"
<returns nothing>
```

Environment data
----------------

OS

- [ ] macOS
- [x] Windows
- [ ] Linux

Server

- [ ] TFS 2017
- [ ] TFS 2018
- [ ] Azure DevOps Server
- [x] Azure DevOps Service

Cannot get anything back for `$VSTeamVersionTable`
However:
```powershell
get-module vsteam -ListAvailable



    Directory: C:\Program Files\WindowsPowerShell\Modules


ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     7.1.3      VSTeam                              {Add-VSTeam, Add-VSTeamAccessControlEntry, Add-VSTeamArea, Add-VSTeamAzureRMServiceEndpoint...}
```

```powershell
> Get-VSTeamAPIVersion
Packaging                      6.0-preview
Version                        VSTS
ServiceEndpoints               5.0-preview
HierarchyQuery                 5.1-preview
Graph                          6.0-preview
Pipelines                      5.1-preview
TaskGroups                     6.0-preview
ExtensionsManagement           6.0-preview
Git                            5.1
VariableGroups                 5.1-preview.1
Core                           5.1
Release                        5.1
DistributedTaskReleased        5.1
MemberEntitlementManagement    6.0-preview
Tfvc                           5.1
Processes                      6.0-preview
Policy                         5.1
Build                          5.1
DistributedTask                6.0-preview
```

```powershell
> $PSVersionTable
PSVersion                      5.1.19041.610
BuildVersion                   10.0.19041.610
```

mnieto:
# PR Summary

Allows  to take advantage of the filter parameter defined in the API v6.0
Defines a new parameterset specific for v6.0 onwards, and ensure that the existing top and skip parameters only can be executed with lower API versions

Fix  #388 

## PR Checklist


Source/Public/Get-VSTeamUserEntitlement.ps1
```diff
@@ -37,23 +73,91 @@ function Get-VSTeamUserEntitlement {
          }
       }
       else {
-         # Build the url to list the teams
+         # use the appropiate syntax depending on the API version
+         $useContinuationToken = $false
```
mnieto: another way instead using the if... else... Not sure what is more readable
```powershell
$useContinuationToken = ($paramCounter -eq 0 -and $apiVersion -gt '6.0') -or ($paramset -contains $PSCmdlet.ParameterSetName)
```

.docs/synopsis/Get-VSTeamUserEntitlement.md
```diff
@@ -1 +1,6 @@
-Get User Entitlement for a user.
\ No newline at end of file
+Get User Entitlement for a user, or a paged list of users matching the specified filter
+
+Please note that Filter, Name, UserType and License parameters only works when MemberEntitlementManagement module version is 6.0 or upper
```
SebastianSchuetze: Synopsis is only for a very short explanation.
Anything more than a sentence should be put only in detailed descriptions. Best is to copy this text directly below the synopsis import in the description part.

.docs/Get-VSTeamUserEntitlement.md
```diff
@@ -14,6 +14,44 @@
```
SebastianSchuetze, line 19: add a new line an put the synopsis text here.

Source/Private/common.ps1
```diff
@@ -121,6 +121,50 @@ function _callAPI {
...
 
+
+function _callAPIContinuationToken {
+   [CmdletBinding()]
+   param(
+      [string]$Url,
+      [switch]$UseHeader,
+      [string]$ContinuationTokenName,
+      [string]$PropertyName,
+      [int]$MaxPages
```
SebastianSchuetze, line 146: Could you please add some description to what effects the properties have within the function?

Source/Private/common.ps1
```diff
@@ -121,6 +121,50 @@ function _callAPI {
    }
 }
 
+
+function _callAPIContinuationToken {
```
SebastianSchuetze, line 130: Please also can you put description what the function is solving in general and a a "TODO:..." that says that this would need to be integrated probably into the _callAPI function?

SebastianSchuetze: Please also add unit test for this function. There is a place to unit test the common functions.

Source/Public/Get-VSTeamUserEntitlement.ps1
```diff
@@ -9,22 +9,57 @@ function Get-VSTeamUserEntitlement {
...
+
+      [Parameter(ParameterSetName = 'PagedFilter')]
+      [Parameter(ParameterSetName = 'PagedParams')]
+      [int] $MaxPages = 0,
+
+      [Parameter(ParameterSetName = 'PagedFilter')]
+      [string] $Filter,
+
+      [Parameter(ParameterSetName = 'PagedParams')]
+      [ValidateSet('Account-Stakeholder', 'Account-Express', 'Account-Advanced', IgnoreCase = $false)]
```
SebastianSchuetze, line 29: I haven't checked yet, but are you sure those are the only valid values? Can you reference a link that states thos? We have to support AzDO Server and Cloud version.

Source/Public/Get-VSTeamUserEntitlement.ps1
```diff
@@ -9,22 +9,57 @@ function Get-VSTeamUserEntitlement {
...
+
    process {
-      # This will throw if this account does not support MemberEntitlementManagement
-      _supportsMemberEntitlementManagement
+      # This will throw if this account does not support MemberEntitlementManagement 
+      # or supported version is not correct with the type of API call
+      $paramCounter = _countParameters -BoundParameters $PSBoundParameters
+      $paramset = 'PagedParams', 'PagedFilter'
+      if ($paramCounter -eq 0) {
+         _supportsMemberEntitlementManagement
```
SebastianSchuetze, line 51: not understanding why you make this difference with the parameter count. IT is either API until 5.1 or above. Why do you need this one?

Source/Public/Get-VSTeamUserEntitlement.ps1
```diff
@@ -37,23 +72,67 @@ function Get-VSTeamUserEntitlement {
...
          }
-
          Write-Output $objs
+
+      }
+   }
+}
+
+
+function _countParameters() {
```
SebastianSchuetze: I would like to avout this function if possible. Firstly, no private functions in public function files. as they belong to the common functions. Secondly, it does look like a very spicific one. Maybe my comment above about the parameter count can lead to, that we don't need this function.

Tests/function/tests/Get-VSTeamUserEntitlement.Tests.ps1
```diff
@@ -32,43 +32,116 @@ Describe "VSTeamUserEntitlement" -Tag 'VSTeamUserEntitlement' {
...
 
             Mock Invoke-RestMethod { Open-SampleFile 'Get-VSTeamUserEntitlement.json' }
             Mock Invoke-RestMethod { Open-SampleFile 'Get-VSTeamUserEntitlement-Id.json' } -ParameterFilter {
                $Uri -like "*00000000-0000-0000-0000-000000000000*"
             }
          }
 
-         It 'no parameters should return users' {
-            $users = Get-VSTeamUserEntitlement
+         Context 'Get-VSTeamUserEntitlement_UpToVersion6.0' {
```
SebastianSchuetze: no underscore. Should be as much readable as possible. Please also do it for the other contexts.

SebastianSchuetze (changes requested):
Really great code as a first contributor!
Not so sure about the solution how you try to check API difference by using count of parameters. Which seems more of a work around.

Wondering if we could solve it by using dynamic parameters and trying to resolve parameters that are shown by evaluating the set apiVersion. The version is always set, and we should make use of it and create the dependency from there.

Source/Public/Get-VSTeamUserEntitlement.ps1
mnieto: Doc source is here: https://docs.microsoft.com/en-us/rest/api/azure/devops/memberentitlementmanagement/user-entitlements/search-user-entitlements?view=azure-devops-rest-6.0

Valid filters include: licenseId, licenseStatus, userType, and name
LicenseId: filters based on license assignment using [license names](https://docs.microsoft.com/en-us/rest/api/azure/devops/memberentitlementmanagement/user-entitlements/search-user-entitlements?view=azure-devops-rest-6.0#accountlicensetype). But license names in the documentation do not match with the values required to build the filter. 

[Here](https://docs.microsoft.com/en-us/azure/devops/organizations/security/access-levels?view=azure-devops#programmatic-mapping-of-access-levels) there is also a description of the license names, but also do not match with the required values. At least in this document Visual Studio subscriptions are excluded as valid license types.

The real license name that apperars in the DevOps UI depends of a combination of accountLicenseType, licensingSource and msdnLicenseType. But the filter paramter in the API only allows filter by licenseId, that, in fact, is not exactly the same that accountLicenseType

It is really a pity not to be able to filter by the license types associated with a VS subscription. And also the documentation is not clear.

Regarding to the AzDO Server, User entlite management calls do not exists. Also in the tests file, [the server version is not supported](https://github.com/MethodsAndPractices/vsteam/blob/41f28b4d9929ed8778e2d001ec3d54d4968cb1aa/Tests/function/tests/Get-VSTeamUserEntitlement.Tests.ps1#L16-L29).

SebastianSchuetze: @mnieto so you are saying those are the only valid licenceId values that work?

Source/Public/Get-VSTeamUserEntitlement.ps1
mnieto: Yes, my first idea was having into account until 5.1 version or above. 

But there is a scenario, when invoking the cmdlet without parameters, that should work on both versions. That is the reason for the 0 parameters check. Then I realized that I used common argumetns like -Verbose, the 0 parameters check fails. And that is the reason I created the _countParameters.

Also, as this function is very specific I put it in the local file instead of moving it the the common functions.

Source/Public/Get-VSTeamUserEntitlement.ps1
mnieto: yes, exactly. I could not find a way to query about VS licenses

Tests/function/tests/Get-VSTeamUserEntitlement.Tests.ps1
mnieto: Ok, I don't know why, I though that spaces was not allowed. Fixed. I will upload de changes sortly

Source/Private/common.ps1
mnieto: Sorry, I didn't realize that common functions had tests. Working on it. Thanks for the remainder

mnieto: Tests added on  commit 235e0da

Source/Public/Get-VSTeamUserEntitlement.ps1
mnieto: Let me know what do you think about my above comment, and I will resolve this one in consequence

mnieto:
@SebastianSchuetze , I sent my comments regarding to the _countParameters function. What direction should I get?

Source/Public/Get-VSTeamUserEntitlement.ps1
SebastianSchuetze: Ok. I leave it like it is here. Currently I don't see a more elegant way.

Source/Public/Get-VSTeamUserEntitlement.ps1
SebastianSchuetze: I moved the function to common.ps1 I think it fits there and is not too special. It could be reused in other functions if needed again.

SebastianSchuetze (approved):

PR #494, opened 2022-11-08
Fix for URL generation with blank api-version

### LINKED ISSUE #442: Get-VSTeamVariableGroup by name throws error 404 with TFS 2020 u1.1
Steps to reproduce
------------------

```powershell
Set-VSTeamAccount "http://tfs.acme.com:8080/tfs/MyCollection/" -UseWindowsAuthentication
Get-VSTeamVariableGroup -ProjectName MyProject -Name MyGroup
```

Expected behavior
-----------------
A group record

Actual behavior
---------------

Error message with a large HTML along the lines of "error 404"

Retrieval by ID works.

If you set the project by `Set-VSTeamDefaultProject`, same result.

Environment data
----------------

OS

- [ ] macOS
- [x] Windows
- [ ] Linux

Server

- [ ] TFS 2017
- [ ] TFS 2018
- [x] Azure DevOps Server
- [ ] Azure DevOps Service

```powershell
Billing                     :
Build                       : 3.0
Core                        : 3.0
DistributedTask             : 3.0-preview
DistributedTaskReleased     :
ExtensionsManagement        : 3.0-preview
Git                         : 3.0
Graph                       :
HierarchyQuery              :
MemberEntitlementManagement :
Packaging                   : 3.0-preview
Pipelines                   :
Policy                      : 3.0
Processes                   :
Release                     : 3.0-preview
ServiceEndpoints            : 3.0-preview
TaskGroups                  : 3.0-preview
Tfvc                        : 3.0
VariableGroups              :
Version                     : TFS2017
```

```
PSVersion                      5.1.18362.1801
BuildVersion                   10.0.18362.1801
```

sevaa:
# PR Summary

fixes #442

## PR Checklist

### DIFF
--- a/Source/Private/common.ps1
+++ b/Source/Private/common.ps1
@@ -375,8 +375,12 @@ function _buildRequestURI {
          $sb.Append("/$id") | Out-Null
       }
 
+      if ($version -or $queryString) {
+         $sb.Append("?") | Out-Null
+      }
+
       if ($version) {
-         $sb.Append("?api-version=$version") | Out-Null
+         $sb.Append("api-version=$version") | Out-Null
       }
 
       $url = $sb.ToString()
--- a/Source/Private/common.ps1
+++ b/Source/Private/common.ps1
@@ -375,21 +375,26 @@ function _buildRequestURI {
          $sb.Append("/$id") | Out-Null
       }
 
-      if ($version -or $queryString) {
-         $sb.Append("?") | Out-Null
-      }
-
+      $QSParams = [System.Collections.ArrayList]@()
       if ($version) {
-         $sb.Append("api-version=$version") | Out-Null
+         $QSParams.Add("api-version=$version") | Out-Null
       }
-
-      $url = $sb.ToString()
 
       if ($queryString) {
          foreach ($key in $queryString.keys) {
-            $Url += _appendQueryString -name $key -value $queryString[$key]
+            $value = $queryString[$key]
+            if($value) {
+               $QSParams.Add("$key=$value") | Out-Null
+            }
          }
       }
+
+      if($QSParams.Count) {
+         $sb.Append("?") | Out-Null
+         $sb.Append($QSParams -join "&") | Out-Null
+      }
+
+      $url = $sb.ToString()
 
       return $url
    }
--- a/Tests/function/tests/Get-VSTeamVariableGroup.Tests.ps1
+++ b/Tests/function/tests/Get-VSTeamVariableGroup.Tests.ps1
@@ -34,6 +34,14 @@ Describe 'VSTeamVariableGroup' {
                $Uri -eq "https://dev.azure.com/test/project/_apis/distributedtask/variablegroups/$($projectID)?api-version=$(_getApiVersion VariableGroups)"
             }
          }
+
+         It 'list should return the variable group with the given name' {
+            Get-VSTeamVariableGroup -projectName project -name theGroup
+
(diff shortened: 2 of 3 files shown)


sevaa:
I didn't touch `Update-VSTeamProject` or its test. Was it broken all along? Can't repro the test fail on Windows.

Source/Private/common.ps1
```diff
@@ -375,18 +375,27 @@ function _buildRequestURI {
          $sb.Append("/$id") | Out-Null
       }
 
+      $QSParams = [System.Collections.ArrayList]@()
```
SebastianSchuetze, line 378: lower case variables only please. Upper case is only for parameters.

Source/Private/common.ps1
```diff
@@ -375,18 +375,27 @@ function _buildRequestURI {
...
+            }
          }
       }
 
+      if($QSParams.Count) {
+         $sb.Append("?") | Out-Null
+         $sb.Append($QSParams -join "&") | Out-Null
+      }
+
+      $url = $sb.ToString()
```
SebastianSchuetze, line 397: Why not using the _appendQueryString and refactor that function?

Somehow it feels like the fix needs a bit more refactoring on also different places to make the code cleaner. Because now we have an approach here to deal with the query string and another with the function (which is not the same as your code but potion of it is trying to do the same).

SebastianSchuetze (changes requested):
Got some comments. Maybe you can give me your take on this.

Source/Private/common.ps1
sevaa: Having a string builder for initial portion of the URL, then growing it by `+=` anyway, is weird to begin with.

Wrapping a `($key, $value) => "$key=$value"` logic into a separate function doesn't make sense to me either :) If it was my project, I'd use a built-in query string generator - .NET RTL has several. For compatibility with .NET 4.x, I'd use `HttpValueCollection`, as constructed via `HttpUtility.ParseQueryString()`. It's in the `System.Web.HttpUtility` assembly though, requires an `Add-Type` call.

What do you think? You are the maintainer, the coding style is your decision after all.

---

OBTW, `_appendQueryString` has another bug, it doesn't URLEncode the values. To reproduce, create a var group called "a&b" and issue the following:

```
Get-VSTeamVariableGroup -ProjectName MyProject -Name "a&b"
```
The .NET query string generator is correct in that regard, just saying :) And unless you want to write your own URLEncode logic, you'd need the `HttpUtility` anyway.

Ref: https://stackoverflow.com/questions/3865975/namevaluecollection-to-url-query

Source/Private/common.ps1
SebastianSchuetze: please change to lower case. I can edit this. Then I will merge.

Source/Private/common.ps1
SebastianSchuetze: I am ok with this. Would you give it a take to use a build it .net query generator for another PR? I would be fine with it if it improves the things. But I would need unit tests for the wrapping function, and it should be replaced / tested everywhere, where it is used.

SebastianSchuetze (changes requested):
Could you please merge the main and update the changelog please? It should be version 7.12.0 now before release.

sevaa:
I'll see what I can do. This PR would be pointless then. Let's keep it alive for the time being, once I have something workable, I'll delete this one.

#496 would be nice to merge though :) I build on PS5.

sevaa:
Covered by #523, closing,

PR #495, opened 2022-11-09
Set-VSTeamVariableGroupVariable cmdlet, tests, docs

### LINKED ISSUE #423: Update Single Variable
There is no cmdlet to update/create a single variable in a variable group. I'd like one, please. There is a task to get and save the whole group, but I'm after a one liner. This is not for me, this is for a community of users, so instructing them to write a multiline powershell script would not be elegant.

That said, a cmdlet like this would go against what seems to be the convention in this module, specifically that each cmdlet wraps a single REST API call and the consumers are expected to do the rest.

What's the best path forward?

1. Leave this issue here, hope the maintainers implement one
2. Implement a cmdlet myself, submit a PR
3. Implement a cmdlet, publish in my own module

Please advice.

sevaa:
# PR Summary

Addresses #423 

## PR Checklist

### DIFF
--- a/.docs/Set-VSTeamVariableGroupVariable.md
+++ b/.docs/Set-VSTeamVariableGroupVariable.md
@@ -1,0 +1,142 @@
+<!-- #include "./common/header.md" -->
+
+# Set-VSTeamVariableGroupVariable
+
+## SYNOPSIS
+
+<!-- #include "./synopsis/Set-VSTeamVariableGroupVariable.md" -->
+
+## SYNTAX
+
+## DESCRIPTION
+
+<!-- #include "./synopsis/Set-VSTeamVariableGroupVariable.md" -->
+
+## EXAMPLES
+
+### Example 1
+
+```powershell
+$methodParameters = @{
+   ProjectName              = "some_project_name"
+   GroupName        = "some_variable_group"
+   Name = "MyVariable"
+   Value = "Foo"
+}
+
+```
+
+## PARAMETERS
+
+### Confirm
+
+Prompts you for confirmation before running the cmdlet.
+
+```yaml
+Type: SwitchParameter
+Aliases: cf
+
+Required: False
+Position: Named
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### Force
+
+Does not prompt
+
+```yaml
+Type: SwitchParameter
+Aliases:
+
+Required: False
+Position: Named
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### WhatIf
+
+Shows what would happen if the cmdlet runs.
+The cmdlet is not run.
+
+```yaml
+Type: SwitchParameter
+Aliases: wi
+
+Required: False
+Position: Named
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### GroupName
+
+Name of the existing variable group
+
+```yaml
+Type: String
+Aliases:
+
+Required: True
+Position: Named
+Default value: None
+Accept pipeline input: True (ByPropertyName)
+Accept wildcard characters: False
+```
+
+### Name
+
+Name of the variable to set
+
+```yaml
+Type: String
+Aliases:
+
+Required: True
+Position: Named
+Default value: None
+Accept pipeline input: True (ByPropertyName)
+Accept wildcard characters: False
+```
+
+### Value
+
+The new value of the variable
+
+```yaml
+Type: String
+Aliases:
+
+Required: False
+Position: Named
+Default value: ""
+Accept pipeline input: True (ByPropertyName)
+Accept wildcard characters: False
+```
+
+<!-- #include "./params/projectName.md" -->
+
+## INPUTS
+
+## OUTPUTS
+
+## NOTES
+
+<!-- #include "./common/prerequisites.md" -->
+
+## RELATED LINKS
+
+
+
+[Add-VSTeamVariableGroup](Add-VSTeamVariableGroup.md)
+
+[Get-VSTeamVariableGroup](Get-VSTeamVariableGroup.md)
+
+[Remove-VSTeamVariableGroup](Remove-VSTeamVariableGroup.md)
+
+[Update-VSTeamVariableGroup](Update-VSTeamVariableGroup.md)
--- a/.docs/Set-VSTeamVariableGroupVariable.md
+++ b/.docs/Set-VSTeamVariableGroupVariable.md
@@ -17,62 +17,14 @@
 ### Example 1
 
 ```powershell
-$methodParameters = @{
-   ProjectName              = "some_project_name"
-   GroupName        = "some_variable_group"
-   Name = "MyVariable"
-   Value = "Foo"
-}
 
+# In project MyProject, variable group MyVarGroup, set the value of Foo to Hello. The variable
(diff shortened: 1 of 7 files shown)


.docs/Set-VSTeamVariableGroupVariable.md
```diff
@@ -0,0 +1,142 @@
...
+
+```powershell
+$methodParameters = @{
+   ProjectName              = "some_project_name"
+   GroupName        = "some_variable_group"
+   Name = "MyVariable"
+   Value = "Foo"
+}
+
+```
```
SebastianSchuetze: The example is missing the actual example call and a short textual description what is happening.

.docs/synopsis/Set-VSTeamVariableGroupVariable.md
```diff
@@ -0,0 +1 @@
+Adds a variable to an existing variable group, or changes its value if the variable is already present. Secret variable updating or creation is not currently supported.
```
SebastianSchuetze: I moved the comment about secret variables to the notes section. More appropiate there.

.docs/Set-VSTeamVariableGroupVariable.md
```diff
@@ -0,0 +1,142 @@
...
+```yaml
+Type: SwitchParameter
+Aliases: wi
+
+Required: False
+Position: Named
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
```
SebastianSchuetze: no need to document them. They are common parameters and will be added automatically when documentation is generated. You only need to document the parameters you declared.

Source/Public/Set-VSTeamVariableGroupVariable.ps1
```diff
@@ -0,0 +1,58 @@
...
+      [Parameter(Mandatory = $true, ValueFromPipelineByPropertyName = $true)]
+      [vsteam_lib.ProjectValidateAttribute($false)]
+      [ArgumentCompleter([vsteam_lib.ProjectCompleter])]
+      [string] $ProjectName
+   )
+
+   process {
+      if ($Force -or $pscmdlet.ShouldProcess("$GroupName.$Name", "Set Variable Value")) {
+         try {
+            $gl = Get-VSTeamVariableGroup -ProjectName $ProjectName
```
SebastianSchuetze, line 24: You are duplicating code here the `Get-VSTeamVariableGroup` has a `-Name`parameter that you should use that eliminates the loop. I know that you are trying to make it more usefule, but this should be handled in the Get-VSTeamVariableGroup. This makes this cmdlet too complicated.

e.g.
`Get-VSTeamVariableGroup -ProjectName $ProjectName -Name $Name`

Source/Public/Set-VSTeamVariableGroupVariable.ps1
```diff
@@ -0,0 +1,58 @@
...
+            # Exception if a group with a name is not found is not particularly descriptive
+            $g = $gl | Where-Object -FilterScript {$_.Name -eq $GroupName}
+            if($g) {
+               if($g.variables | Get-Member -Name $Name) {
+                  if(($g.variables.$Name | Get-Member -Name isSecret) -and $g.variables.$Name.isSecret) {
+                     Write-Error "The variable $Name is a secret one. Updating secret variables is not currently supported."
```
SebastianSchuetze, line 30: Good one. Makes it easier to understand!

Source/Public/Set-VSTeamVariableGroupVariable.ps1
```diff
@@ -0,0 +1,58 @@
...
+                      $g.variables.$Name.value = $Value
+                  }
+               }
+               else {
+                  $Vars = @{}
+                  foreach($v in $g.variables.PSObject.Properties) {
+                     $Vars[$v.Name] = $v.Value
+                  }
+                  $Vars[$Name] = @{value=$Value}
+                  $g.variables = New-Object PSObject -Property $Vars
```
SebastianSchuetze: Instead of recreating the whole object, rather use the Add-Member cmdlet to add a new object to the existing varaibles:

`Add-Member -InputObject $g.variables -MemberType NoteProperty -Name test -Value ([pscustomobject]@{value='val1'})`

Easier to understand and less code.

Source/Public/Set-VSTeamVariableGroupVariable.ps1
```diff
@@ -0,0 +1,58 @@
...
+               }
+               $Body = $g | ConvertTo-Json -Depth 20 -Compress
```
SebastianSchuetze: lowercase please. Pascal Casing is only for parameters and not variables.

Tests/function/tests/Set-VSTeamVariableGroupVariable.Tests.ps1
```diff
@@ -0,0 +1,54 @@
...
+            Mock _getInstance { return 'https://dev.azure.com/test' }
+            Mock Get-VSTeamVariableGroup { @(@{
+               name = "TheGroup";
+               id = 101;
+               variables = @{
+                  "x" = @{value="A"};
+                  "y" = @{value="B"};
+                  "s" = @{value=$null; isSecret=$True}
+               }
+            }) | ConvertTo-JSON | ConvertFrom-JSON}
```
SebastianSchuetze, line 21: Why are you doing a double convertion here?

SebastianSchuetze (changes requested):
Thanks for the contribution. I added my comments to the code lines. Just some minor things to make the code cleaner and more maintainable in the future.

.docs/Set-VSTeamVariableGroupVariable.md
sevaa: Sorry, copy&paste from the doc file for `Update-VSTeamVariableGroup`.

Source/Public/Set-VSTeamVariableGroupVariable.ps1
sevaa: `Get-VSTeamVariableGroup` throws three exceptions if the -Name corresponds to a nonexistent group, and the exception text is rather misleading ("You cannot call a method on a null-valued expression."). If you'd like me to address that separately, let me know.

For now, I'm afraid relying on Get...Group with -Name would be a user unfriendly thing to do.

Tests/function/tests/Set-VSTeamVariableGroupVariable.Tests.ps1
sevaa: So that the output of the mocked Get...Group is PSObjects instead of hashtables, like the real Get...Group provides.

sevaa:
Addressed most. Please take another look.

sevaa:
What are my next steps, please?

The group search by name is broken (obscure exception if not found), the double JSON convert can be avoided but the review doesn't say so, just asks why.

sevaa:
Is this PR a lost cause? If so, please let me know. As outlined in #423, I'll go ahead and publish as a separate PowerShell module.

sevaa:
Any update please? Any maintainers present? I think I've made all the changes needed - but the PR still says "Changes requested". @SebastianSchuetze @DarqueWarrior

SebastianSchuetze (dismissed):

sevaa:
@SebastianSchuetze What should be my actions now?

SebastianSchuetze:
Thanks for the help! Gonna merge it.

Issue #412, opened 2021-08-05
Get-VSTeamUser return only 500 users event if there is more, Filtering by Get-VSTeamUser  | ? DisplayName also not finding existing user


JanuszNowak:
Steps to reproduce
------------------

```powershell

$users = Get-VSTeamUser 

```

Expected behavior
-----------------

```none
$users = Get-VSTeamUser 
$users | Measure
2537
```

Actual behavior
---------------

```none
$users = Get-VSTeamUser 
$users | Measure
500
```

Environment data
----------------

OS

- [ ] macOS
- [x ] Windows
- [ ] Linux

Server

- [ ] TFS 2017
- [ ] TFS 2018
- [ ] Azure DevOps Server
- [ x] Azure DevOps Service

```powershell
> Get-VSTeamAPIVersion
Name                           Value                                                                                                                                                  
----                           -----                                                                                                                                                  
Packaging                      6.0-preview                                                                                                                                            
Version                        VSTS                                                                                                                                                   
ServiceEndpoints               5.0-preview                                                                                                                                            
HierarchyQuery                 5.1-preview                                                                                                                                            
Graph                          6.0-preview                                                                                                                                            
Pipelines                      5.1-preview                                                                                                                                            
TaskGroups                     6.0-preview                                                                                                                                            
ExtensionsManagement           6.0-preview                                                                                                                                            
Git                            5.1                                                                                                                                                    
VariableGroups                 5.1-preview.1                                                                                                                                          
Core                           5.1                                                                                                                                                    
Release                        5.1                                                                                                                                                    
DistributedTaskReleased        5.1                                                                                                                                                    
MemberEntitlementManagement    6.0-preview                                                                                                                                            
Tfvc                           5.1                                                                                                                                                    
Processes                      6.0-preview                                                                                                                                            
Policy                         5.1                                                                                                                                                    
Build                          5.1                                                                                                                                                    
DistributedTask                6.0-preview
```

```powershell
> $PSVersionTable
Name                           Value                                                                                                                                                  
----                           -----                                                                                                                                                  
PSVersion                      5.1.19041.1151                                                                                                                                         
PSEdition                      Desktop                                                                                                                                                
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                
BuildVersion                   10.0.19041.1151                                                                                                                                        
CLRVersion                     4.0.30319.42000                                                                                                                                        
WSManStackVersion              3.0                                                                                                                                                    
[...]

Issue #169, opened 2019-06-18
Get more than 100 Releases


vercellone:
_Originally posted by @stefanstranger in https://github.com/DarqueWarrior/vsteam/issues/35

The REST API includes an x-ms-continuationtoken response header that, unfortunately, is not readily accessible when using Invoke-RestMethod.  In order to access the response header(s) it may be necessary to use Invoke-WebRequest (or similar).

Steps to reproduce
------------------
```powershell
Get-VSTeamRelease -definitionId 12 -Top 200
Get-VSTeamRelease -definitionId 12 -ContinuationToken '???'
```

Expected behavior
-----------------
- -Top 200 should return 200
- Results should always include a ContinuationToken (if the response headers include x-ms-continuationtoken) to use in subsequent calls in order to retrieve all releases.

Actual behavior
---------------
- -Top 200 returns 100
- No ContinuationToken is returned, therefore it is not possible to request another page.

Environment data
----------------
OS
- [ ] macOS
- [X] Windows
- [ ] Linux

Server
- [ ] TFS 2017
- [ ] TFS 2018
- [ ] Azure DevOps Server
- [X] Azure DevOps Service

```powershell
> Get-VSTeamAPIVersion
Name                           Value                                                                                                                                                                         
----                           -----                                                                                                                                                                         
ServiceFabricEndpoint          5.0-preview                                                                                                                                                                   
Build                          5.0                                                                                                                                                                           
ExtensionsManagement           5.1-preview                                                                                                                                                                   
Release                        5.1-preview                                                                                                                                                                   
Version                        VSTS                                                                                                                                                                          
MemberEntitlementManagement    5.1-preview                                                                                                                                                                   
Packaging                      5.1-preview                                                                                                                                                                   
Tfvc                           5.0                                                                                                                                                                           
Git                            5.1-preview                                                                                                                                                                   
Core                           5.0                                                                                                                                                                           
Graph                          5.1-preview                                                                                                                                                                   
DistributedTask                5.0-preview  
```

```powershell
> $PSVersionTable
Name                           Value                                                                                                                                                                         
----                           -----                                                                                                                                                                         
PSVersion                      5.1.17134.765                                                                                                                                                                 
PSEdition                      Desktop                                                                                                                                                                       
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                                       
BuildVersion                   10.0.17134.765                                                                                                                                                                
[...]

PR #509, opened 2023-01-06
Implement continuationToken management (body and headers) on v8.0.0 branch

mnieto:
# PR Summary

Some endpoints in Azure DevOps APIs version 6.0 and above use continuationToken as part of the body response or X-MS-ContinuationToken header

In both cases it is possible provide this functionality adding the following parameters to the __callAPI_ function
*  **[string]$UseContinuationToken = 'None'** Posible values: None, Header, Body
* **[string]$ContinuationTokenName**  Allows to specify a header or continuation token property different of the default values. If this parameter is not specified, the default value is X-MS-ContinuationToken or continuationToken depending if $UseHeader is present or not, respectively. Ignored if $UseContinuationToken -eq 'None'
* **[int]$MaxPages = 0**   Number of pages to be retrieved. If 0, or not specified, it will return all the available pages. Ignored if $UseContinuationToken -eq 'None'
* **[string]$ColectionPropertyName**  When using continuationToken, it is neccesary expand a specific property to get the real  collection of objects. Ignored if $UseContinuationToken -eq 'None'

**IMPORTANT**: Managing response headers in the `Invoke-RestMethod` cmdlet forces to use only poweshell core. This required functionality was introduced in [powershell core v6](PowerShell/PowerShell#4888) and it is not available in Desktop flavour.

With this changes it is not necessary to maintain a separated ___callAPIContinuationToken_ function. As part of the PR I want to update the `Get-VSTeamUser` and `Get-VSTeamUserEntitlement` functions mentioned in the below issues.

This change enables the possibility to fix at least the below issues:
* #412
* #388 Already fixed with a partial implementation
* #169 

## PR Checklist

### DIFF
--- a/.docs/Get-VSTeamUserEntitlement.md
+++ b/.docs/Get-VSTeamUserEntitlement.md
@@ -6,6 +6,11 @@
 
 <!-- #include "./synopsis/Get-VSTeamUserEntitlement.md" -->
 
+Please note that Filter, Name, UserType and License parameters only works when MemberEntitlementManagement module version is 6.0 or upper
+In the same way Top and Skip paramerers only works up to version 5.1
+
+You can setup the specific version for the MemberEntitlementManagement calling Set-VSTeamAPIVersion -Service MemberEntitlementManagement -Version VersionNumberYouNeed.
+
 ## SYNTAX
 
 ## DESCRIPTION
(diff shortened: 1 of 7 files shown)


mnieto:
@SebastianSchuetze, perhaps it could be a good idea to announce, in addition to #458 the removal of support to Powershell Desktop in the v8.x version

PR #532, opened 2023-05-22
fix #531 when Version is not set in Set-VSTeamAccount

### LINKED ISSUE #531: Set-VSTeamAccount Error on Module Import
### Steps to reproduce

```powershell
Import-Module -Name VSTeam
```
I'm getting this error while importing the module on version `7.13.0`. Normally, I wouldn't even use the `-Version` parameter when using `Set-VSTeamAccount` cmdlet.

### Expected behavior

The module would import without errors.

### Actual behavior?

I see the following error displayed:

```text
Set-VSTeamAccount: Cannot validate argument on parameter 'Version'. The argument "" does not belong to the set "TFS2017,TFS2018,AzD2019,VSTS,AzD,TFS2017U1,TFS2017U2,TFS2017U3,TFS2018U1,TFS2018U2,TFS2018U3,AzD2019U1" specified by the ValidateSet attribute. Supply an argument that is in the set and then try the command again.
```

### On Which OS have you tried it?

Windows

### What was your server version?

Azure DevOps Services

### Other server version

_No response_

### Log output of used API

```shell
Billing                     :
Build                       : 3.0
Core                        : 3.0
DistributedTask             : 3.0-preview
DistributedTaskReleased     :
ExtensionsManagement        : 3.0-preview
Git                         : 3.0
Graph                       :
HierarchyQuery              :
MemberEntitlementManagement :
Packaging                   : 3.0-preview
Pipelines                   :
Policy                      : 3.0
Processes                   :
Release                     : 3.0-preview
ServiceEndpoints            : 3.0-preview
TaskGroups                  : 3.0-preview
Tfvc                        : 3.0
VariableGroups              :
Version                     : TFS2017
Wiki                        :
WorkItemTracking            : 3.0
```

### Log output of $PSVersionTable

```shell
PSVersion                      7.3.4
PSEdition                      Core
GitCommitId                    7.3.4
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
```

mnieto:
# PR Summary

fix #531 when Version is not set in Set-VSTeamAccount

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 7.13.1
+Merged [Pull Request](https://github.com/MethodsAndPractices/vsteam/pull/532) from [Miguel Nieto](https://github.com/mnieto) the following:
+- Fix Set-VSTeamAccount Error on Module Import [531](https://github.com/MethodsAndPractices/vsteam/issues/531)
+
+
 ## 7.13.0
 
 Merged [Pull Request](https://github.com/MethodsAndPractices/vsteam/pull/515) from [mrwalters1988](https://github.com/mrwalters1988) the following:
--- a/Source/VSTeam.psm1
+++ b/Source/VSTeam.psm1
@@ -38,7 +38,9 @@ if ($null -ne $env:TEAM_PROJECT) {
       # set vsteam account to initialize given variables properly
       $commonArgs = @{
          Account = $env:TEAM_ACCT
-         Version = $env:TEAM_VERSION
+      }
+      if (-not [string]::IsNullOrEmpty($env:TEAM_VERSION)) {
+         $commonArgs["Version"] = $env:TEAM_VERSION
       }
 
       if (_useBearerToken) {


SebastianSchuetze:
Can you explain what happens if the vsteam module is already set and it's still overwritten.

I am thinking to start to write unit tests for the psm1 with Pester. This seems to get more complicated. So I think we should test the import part as well.

SebastianSchuetze:
Ah I see the issue

mnieto:
> I am thinking to start to write unit tests for the psm1 with Pester. This seems to get more complicated. So I think we should test the import part as well.

Yes, I agree
I don't have experience enough with Pester.  .psm1 files are complex to test. Perhaps we can move most of the functionality to an internal testable function and keep the psm1 file as much simple as possible

SebastianSchuetze:
There is a recommendation:

https://pester.dev/docs/usage/modules

Testing the functions would be fine but also testing the import on different OS itself would also be important. I have done this at work but I got it from an example on github. I need to check.

SebastianSchuetze:
Found it: https://powershellexplained.com/2017-01-21-powershell-module-continious-delivery-pipeline/#psm1-module

